PR #15, opened 2026-05-27
security(download_aw): verify SHA-256 of bundled ActivityWatch download + use mkstemp

BetterQAFTW:
## What & why

`scripts/download_aw.py` fetches the bundled ActivityWatch tracker zip — the binaries we ship as our own trusted trackers — with **no integrity check** and into a path from `tempfile.mktemp()`:

- `urllib.request.urlretrieve(url, tmp)` with no checksum: a tampered GitHub release asset, a compromised mirror, or a MITM on the download would be extracted and shipped as `bf-data-service` / `bf-window-tracker` / `bf-idle-tracker`.
- `tempfile.mktemp()` returns a name without creating the file — a classic TOCTOU race / symlink-attack window between name generation and write.

## Change

- **Pin SHA-256 per platform asset** (`EXPECTED_SHA256`, keyed by `AW_VERSION`) and verify the download against it. Abort with exit 1 on mismatch, **and** abort if no digest is pinned for the asset (so we never extract an unverified binary). Bump procedure documented inline.
- **Replace `mktemp()` with `mkstemp()`** — atomic `O_EXCL` create, no race. The fd is closed before `urlretrieve` writes the path.
- **Clean up the temp file on any failure/interrupt** before re-raising.

## Verification

- `python3 -c "import ast,sys; ast.parse(open('scripts/download_aw.py').read())"` → parses clean.
- Digests computed from the canonical v0.13.2 GitHub release assets (`shasum -a 256`) and **verified end-to-end against live downloads** of all three zips (macOS / Windows / Linux x86_64) — all MATCH; every platform in `RELEASE_ASSETS` has a pinned digest.
- Negative test: an unpinned asset aborts with exit 1; confirmed `mktemp(` is gone and `mkstemp(` is in use.


BetterQAFTW:
ci(security): added a gitleaks secret-scanning gate (`.github/workflows/gitleaks.yml` + `.gitleaks.toml`).

- Runs `gitleaks/gitleaks-action@v2` on push + pull_request with full-history scan (`fetch-depth: 0`), failing the build on any detected secret.
- `.gitleaks.toml` extends the default ruleset and allowlists known example/placeholder values (`.env.example`, `your_*_here`, `change_me`, public anon JWTs, lockfiles/fixtures) to avoid false positives.

This is the control that would have caught the committed secrets removed on this branch.

tudorsss:
Closing — the runtime vulnerability this targets is already fixed, and the rest of the diff has aged out.

The path that matters, `_download_aw_binaries` in `src/aw_manager.py:296-344`, is the one that downloads and executes the tracker on fleet machines. It already verifies SHA-256 against a pinned `RELEASE_SHA256` dict and fails closed on mismatch, uses `tempfile.mkstemp`, enforces an HTTPS + GitHub-host allowlist, re-checks the host after redirect, and caps the download at 500MB. That landed independently as `790b7a1` (2026-07-18), part of a chain of nine related security commits.

What is still uncovered is `scripts/download_aw.py`, the build-time script that bundles binaries on the CI runner. It never runs on an employee machine, so it is a genuinely lower-severity gap — but it is a real one, and it is the only security-relevant file in this diff.

The blocker to landing this is size rather than merit: 52 files, +1856/-896, mostly a codebase-wide formatting pass, plus dependabot and gitleaks config. It is CONFLICTING with all checks failing, and 33 commits have touched these files since May, so this is a rewrite rather than a rebase.

Better path: a small fresh PR porting just the checksum + mkstemp change to `scripts/download_aw.py` against current `main`. The dependabot and gitleaks additions are independently useful and could ship on their own.

PR #88, opened 2026-06-25
fix(afk): salvage the observed active tail across a long pause/sleep

tudorsss:
## The bug

Reported in Slack by Ecaterina Cocora for **Matei Cocora's device (id 44, agent 1.5.74)** on 2026-06-25: *"some kind of 20-minute issue either before a pause or after one."* A ~10–20 min **idle/empty gap hugs each pause** in the Activity Timeline.

## Root cause (confirmed with real data)

`agent:debug-day --device=44 --date=2026-06-24` shows the "pauses" are **macOS sleep events** — two of them (2.1h and 4.1h) exceed the agent's in-process AFK **sample retention (2h / 7200s)**.

When a sleep is longer than retention, the checkpoint freezes past the retention window and `_build_inproc_afk` re-seeds. The old re-seed **discarded the entire `[checkpoint, now]` window** — including the genuine active work the agent had already observed in the final minutes before the lid closed. With no not-afk evidence there, the server renders those slots as an idle/empty gap (and that time isn't counted).

The committed/deployed **server** timeline can't itself open a mid-day gap (it fills from the window stream), so the missing evidence is upstream in the agent — this is the same family as the still-open AFK upload-stream-gap convergence work.

## The fix (agent-side, two parts)

- **`AfkSource.record_sample(protect_since=…)`** — keep samples at/after a protected floor (the engine's checkpoint) from being pruned, so the active samples taken just before a long pause survive the wake-cycle prune.
- **`_build_inproc_afk` re-seed branch** — split the `now < cp` (clock-backward) case from the `gap > retention` (pause) case. For a pause it now calls `build_afk_events(…, cover_unsampled=False)` to **salvage only the spans the retained samples genuinely cover**, leaving the unobserved sleep window uncovered (never billed active), then re-seeds to `now`. The clock-backward path is unchanged.

`cover_unsampled=False` is new on `build_afk_events`: it omits the unsampled boundary regions instead of filling them with afk, so we never invent activity over a genuinely unobserved window (the original finding-C concern stays honoured).

## Test (proof of failure pre-fix)

`tests/test_inproc_afk_pause_salvage.py`:
- `test_long_pause_reseed_salvages_observed_active_tail` — **fails pre-fix** (re-seed returned `[]`, the pre-pause not-afk span was lost); **passes post-fix** (the observed active tail is emitted as not-afk while the unobserved window stays uncovered).
- `test_long_pause_with_no_observed_tail_emits_nothing` — guards that a blank window still emits nothing.

Full suite: **769 passed**. Existing hardening tests (findings A–F, incl. `gap>retention` with no tail and clock-backward re-seed) pass unchanged.

## Scope / follow-up

- Needs a fleet bump (1.5.7x) to take effect.
- The server still renders the genuinely-unobserved sleep window as a gap; making that read as gray idle rather than white is a separate `internal-tool2` display change.

### DIFF
--- a/src/sync/afk_source.py
+++ b/src/sync/afk_source.py
@@ -29,6 +29,7 @@ class AfkSource:
         input_watcher=None,
         idle_clock: Callable[[], Optional[float]] = get_system_idle_seconds,
         retention_seconds: float = 7200.0,
+        max_samples: int = 20000,
     ) -> None:
         self._afk_timeout = float(afk_timeout_seconds)
         self._hostname = hostname
@@ -37,8 +38,14 @@ class AfkSource:
         self._retention_seconds = float(retention_seconds)
         self._retention = timedelta(seconds=retention_seconds)
         self._lock = threading.Lock()
-        # (sample_time, last_input_at)
-        self._samples: deque = deque()
+        # (sample_time, last_input_at). `maxlen` is an absolute memory backstop:
+        # `protect_since` can hold samples past the retention prune (so a long
+        # pause's pre-pause tail survives, and a held checkpoint can rebuild),
+        # but a pathological frozen checkpoint during a multi-day outage must
+        # not grow this without bound. ~20k samples ≈ 2 weeks at one per cycle;
+        # appends auto-evict the oldest beyond that (salvage past that horizon is
+        # moot anyway).
+        self._samples: deque = deque(maxlen=max_samples)
         # Sticky platform-capability latch: a transient idle-clock read failure
         # (e.g. a one-off `ioreg` timeout) must NOT revoke in-process mode for a
         # cycle — that would hand billing back to the external tracker and flap
@@ -183,6 +190,8 @@ class AfkSource:
                 spans.append((range_start, range_end, "afk"))
         else:
             first = activity[0]
+            # `first > range_start` also covers anchor == range_start (checkpoint
+            # set exactly to a last_input): no leading gap, so skip the span.
             if first > range_start and cover_unsampled:
                 spans.append((range_start, first, "afk"))  # leading unknown
             for a, b in zip(activity, activity[1:], strict=False):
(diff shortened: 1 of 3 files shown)


tudorsss:
## Review — ✅ ready (no critical issues)

Independent review pass (senior-code-reviewer agent + manual analysis). Verdict: **no over-billing bug**, and the in-process-AFK audit invariants are preserved:
- **C** (no reconstruction over unobserved windows) — held: `cover_unsampled=False` drops the unsampled leading/trailing regions, so the sleep window is never filled with inferred activity.
- **D** (clock-backward re-seed) — unchanged, now its own arm.
- **E** (blind-clock hold) — unchanged, still after the salvage branch.
- **F** (ms-precision ids) — salvaged spans can't collide with the wake cycle's ids.

Walked the `build_afk_events(cover_unsampled=False)` span logic: the only paths that could emit a `not-afk` (billable) span over an unobserved window are the leading/trailing boundary spans, and after a >retention sleep `finalize_point` forces the trailing span to `afk` (which is then dropped). Emitted not-afk spans are exclusively inter-sample gaps from directly-observed pre-pause samples. No over-bill.

### Addressed in the follow-up commit (dadb3d8)
- **Memory backstop**: `protect_since` is also needed for finding-B rebuild (a held checkpoint must retain samples), so it can't be capped by time. Added an absolute `deque(maxlen=20000)` so a pathological frozen checkpoint during a multi-day outage can't grow the buffer without bound. New test `test_protect_since_respects_maxlen_backstop`.
- **Checkpoint read race** (the new lock-free `protect_since` read): documented as benign — GIL-atomic, `_advance_checkpoints_to_now` only moves it forward, and `_build_inproc_afk` re-reads it fresh, so a race only over-retains a few samples (bounded by maxlen), never mis-bills. Added a lock-free hot-path read instead of a lock by design.
- **Finding-B deviation** in the salvage branch documented (queue is the durability layer; salvaged spans can't be rebuilt next cycle).
- Clarified the `anchor == range_start` guard; annotated the no-observed-tail test as a guard, not a pre-fix regression test.

Full suite **770 passed**, ruff clean on changed files.

PR #97, opened 2026-06-29
feat: stop collecting outside enforced working hours (capture gate)

BetterQAFTW:
## Why

The desktop agent had no concept of working hours: it recorded window titles and browser URLs **24/7** whenever the machine was active, then relied on the server to *exclude* out-of-hours activity from billed totals ("recorded but not counted"). For a restricted user (B2E / Trainee), that meant their personal evening/weekend activity was still captured, transmitted, and stored where an admin could see it — only filtered after the fact.

This makes the agent **stop collecting** outside enforced hours, rather than merely declining to upload.

## What

- **Outside enforced hours, capture is fully suspended:** the window-title watcher (in-process on macOS, the `bf-window-tracker` subprocess elsewhere) and the browser-URL poller are stopped, and the sync engine is paused. Nothing personal is read, recorded, or sent.
- **Fail-closed at launch:** a returning restricted user who opens their laptop outside hours never starts the watchers — the decision is made from the persisted schedule *before* the first server call, then reconciled the moment the server config lands after login. (Fixes a latent bug where `working_hours` wasn't restored from disk at all.)
- **Auto-resumes at the work-window start:** the 60s tick re-evaluates the gate (before the idle check), so capture restarts within a tick when the clock crosses into the window — no user action.
- **"Unless strictly requested":** a **"Work outside hours (today)"** tray item arms a one-day override that auto-expires at local midnight; shown only to a restricted user currently outside their window.
- **Backstop kept:** the existing upload-side filter remains, refactored to share the single window-check function so the gate and the filter can never disagree.
- **Unrestricted (B2B / flexible) users are unaffected** — they still record 24/7. The gate only bites `enforced` schedules.

## Commits

1. `fix(config)` — restore `working_hours` on load + window-check helpers
2. `feat(schedule)` — `WorkScheduleGate` decision module + tests
3. `feat(aw)` — `suspend_watcher`/`resume_watcher`
4. `fix(window-watcher)` — clear stop-event on `start()` so it's restartable
5. `refactor(sync)` — share window check in upload filter + `set_browser_tracker`
6. `feat(tray)` — "Work outside hours" override item + schedule state
7. `feat(main)` — wire the gate into tick / launch / suspend-resume

## Testing

- **841 tests pass** (16 new: gate decision/boundaries/override expiry, config save→load round-trip, tray wiring). New code is `ruff`-clean.
- Thread-safe (gate override + `_schedule_suspended` are locked); every exception handler logs; window-check logic deduplicated into `WorkingHoursConfig.is_within_window`.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -1,4 +1,4 @@
 """BetterFlow - ActivityWatch to BetterFlow sync companion app."""
 
-__version__ = "1.5.85-beta.2"
+__version__ = "1.5.85-beta.3"
 __author__ = "BetterQA"
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -1,4 +1,4 @@
 """BetterFlow - ActivityWatch to BetterFlow sync companion app."""
 
-__version__ = "1.5.85-beta.3"
+__version__ = "1.5.85-beta.4"
 __author__ = "BetterQA"
--- a/src/aw_manager.py
+++ b/src/aw_manager.py
@@ -473,6 +473,57 @@ class AWManager:
         with self._lifecycle_lock:
             self._disabled_components.add(name)
 
+    def suspend_watcher(self, name: str) -> None:
+        """Stop a managed watcher AND disable it so the watchdog won't revive it.
+
+        Used by the working-hours capture gate to stop the subprocess window
+        tracker (Windows/Linux) outside enforced hours, so no window titles are
+        recorded even locally. Disabling first means restart_if_needed and
+        _start_locked both skip it (they honour _disabled_components), so the
+        terminate below can't race a restart. No-op on macOS where the
+        subprocess tracker is already disabled in favour of the in-process
+        watcher. Mirrors _stop_idle_tracker_locked; reaps orphans so a killed
+        process can't keep posting to the bucket."""
+        with self._lifecycle_lock:
+            self._disabled_components.add(name)
+            proc = self._processes.pop(name, None)
+            if proc is not None and proc.poll() is None:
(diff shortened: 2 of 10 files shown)


tudorsss:
Closing as **superseded**, not rejected — the behaviour shipped, under different names.

Verified on `origin/main` (`058fa42`) rather than inferred from the tracker:

- `src/main.py:2350/2358` calls `set_capture_suppressed(False/True, reason)` and starts/stops watchers on the working-hours decision.
- `src/aw_manager.py:846` — *"Tracker start refused: capture is suppressed"* — and every restart path (`:1055`, `:1083`, `:1209`) is gated, so a restart cannot quietly resume recording.
- The launch-restore bug this PR named is fixed: `src/config.py:962` documents `working_hours` having been missing from the restore list "until 2026-07-14".
- The tray carries a `PRIVATE_HOURS` state, *"outside working hours, not recording"*.

It landed via **#133** (merged 2026-07-14), refined by #134/#135/#136. My first probe searched for this PR's own symbol names and returned zero — which reads as "not shipped" and is exactly wrong. The behaviour was there under other names.

**Fail-closed is stronger here than this PR proposed:** an unknown schedule now refuses to record until the server sends one, rather than defaulting to recording.

**One sub-feature did not land** — the *"Work outside hours (today)"* one-day tray override. Confirmed absent from `src/ui/tray.py`, with a positive control (59 `Item(` matches in the same file, so the probe was live). If that is still wanted it deserves its own issue rather than keeping this branch open.

Closed on Brad's instruction after a status sweep.

PR #112, opened 2026-07-01
fix(queue): drain around a stuck head instead of freezing the queue

BetterQAFTW:
## Why

Follow-up to the 06-30 queue-durability train (#98–#104). Those fixes stopped a
transient outage from **dropping** good activity, but left a silent **liveness**
hole: a batch that fails transiently but *deterministically* (a content-specific
5xx, or a server stuck returning no-confirmation) sat at the oldest-first dequeue
head forever, **freezing every newer event** up to the 30-day expiry — with the
heartbeat green the whole time.

Confirmed with a red-first test: only the poison head was ever attempted; all
newer activity stayed frozen.

## What

Three small, individually-green commits:

1. **refactor(queue): single source of truth for the 7-day retention window** —
   extract the duplicated `7` into `STALE_AFTER_DAYS`, documented as lockstep
   with the server cutoff (`AgentEventProcessor::isValidEventTimestamp`, which
   rejects events >7d old). No behaviour change.
2. **feat(queue): add `dequeue(offset=…)`** — skip the oldest N eligible events
   under a stable total order (`created_at, id`). Additive primitive, unit-tested.
3. **fix(queue): drain around a stuck head instead of freezing the queue** — the
   drain no longer `break`s on failure; it steps past the stuck head with a
   growing offset so the events **behind** it still get served. In-memory,
   per-cycle, so a recovered server drains immediately (no persisted backoff to
   wait out). Bails after N consecutive transient failures so a real outage backs
   off instead of hammering.

### Guarantees preserved
- A transient outage still drops **nothing** (#99).
- A definitive 4xx still increments `retry_count` and drops after `max_retries`.
- `remove_failed` and `expire_old` remain the only delete paths.

### Removed
- The stuck-head-ceiling shed heuristic (`_STUCK_HEAD_CEILING` /
  `_batch_has_storable_activity`). The offset drain makes shedding-to-unblock
  unnecessary, so we no longer need a ~3h timer that could drop good recent
  activity.

## Tests
- Red-first `test_recent_poison_head_does_not_freeze_newer_events` (fails on
  main, passes here).
- `dequeue(offset=…)` unit tests; full-outage-holds-all; storable-head-held;
  definitive-drop retained.
- Full suite: **840 passed** locally.

## Note for review
This reworks logic @tuuude just landed across #98–#104 (4 audit rounds), so an
extra set of eyes from that context would be ideal. Draft until reviewed.

### DIFF
--- a/src/sync/queue.py
+++ b/src/sync/queue.py
@@ -294,14 +294,28 @@ class OfflineQueue:
             )
             return count
 
-    def dequeue(self, batch_size: int = 100, max_retries: int = 5) -> list[QueuedEvent]:
+    def dequeue(
+        self,
+        batch_size: int = 100,
+        max_retries: int = 5,
+        *,
+        offset: int = 0,
+    ) -> list[QueuedEvent]:
         """Get a batch of events from the queue (oldest first).
 
         Skips events that have exceeded max_retries (N7).
 
+        ``offset`` skips the oldest N eligible events. _process_queue uses it to
+        step past a batch that just failed this cycle (which stays at the head
+        because it isn't removed), so the events BEHIND a stuck head still get
+        served instead of being head-of-line-blocked. Ordering is a stable total
+        order (created_at, then id) so a fixed offset is deterministic even when
+        several events share a created_at.
+
         Args:
             batch_size: Maximum number of events to return
             max_retries: Skip events with retry_count >= this value
+            offset: Number of oldest eligible events to skip
 
         Returns:
             List of QueuedEvent objects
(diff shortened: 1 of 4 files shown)


tudorsss:
Closing — merging this as-is would revert a shipped fix.

This PR and #129 solve the same bug. The symptom described here (a content-specific 5xx sits at the head of the queue forever while the heartbeat stays green) is the same one #129 addressed, and #129 merged on 2026-07-10 as `7650708` — nine days after this branch was opened. This diff still reads as authored against pre-#129 `main`.

The problem is that the two solutions are incompatible in a way that is easy to miss. This PR **deletes** `_STUCK_HEAD_CEILING` and `_batch_has_storable_activity` and replaces them with an offset-based dequeue skip-around. Those are exactly the mechanism #129 shipped, so merging this would remove the live fix and leave two approaches racing to decide what happens to a stuck head.

There is a second-order effect worth recording, because neither PR references the other. #131 adds `requeue_storable_dead_letter()`, which replays rows that `_STUCK_HEAD_CEILING` feeds into `dead_letter_events`. If this PR merged, that ceiling would be gone and #131's replay would have nothing to replay — the feature would ship silently inert. Merging both would be worse than merging neither.

If the offset-based approach is still considered better than what #129 shipped, that is worth having as its own discussion with a fresh diff against current `main`. It is not a rebase.

PR #114, opened 2026-07-01
fix(sync): coalesce same-app window title-flicker instead of dropping it

BetterQAFTW:
## Why

Diagnosed from Sachi Navodi's agent log (2026-07-01, Windows, 1.5.87). Her
dashboard showed **"No activity tracked"** while hours still accrued. The log
says it outright, every cycle:

> *Window/app data has gone stale on the server across 3 cycles — the watcher IS
> producing window events but the filter is dropping them all (99 window event(s)
> under the 5s minimum (flicker filter)). Billing is unaffected (AFK/input still
> upload); only per-app attribution is lost.*

Her window title updates faster than `min_window_event_seconds` (chrome.exe with
a live-updating title), so AW emits a new sub-5s window event on every title
change and `_transform_event` drops **all** of them — the per-app timeline goes
empty even though she's continuously in one app.

(Her other reported "no data" was a **separate**, already-resolved issue: an
06-30 backend HTTP 500 storm — 822 errors, all on 06-30 — not the agent, and not
the queue.)

## What

`_coalesce_window_flicker`: merge a run of consecutive, contiguous, **same-app**
window events into one span *before* the minimum is applied, so continuous focus
survives a flickering title. Mirrors the trusted `_collapse_afk_duplicates`
pattern.

- **Gated**: only merges runs that would otherwise lose data (≥2 events, at least
  one below the minimum). A normal user's already-passing events are
  byte-identical — strictly additive.
- **Different apps never merge**, so genuine cross-app alt-tab noise is still
  suppressed by the per-event minimum.
- Representative title/url = the longest-lived event in the run.

## Safety
- Window/web events are **attribution-only** (billing rides the AFK/input
  stream), and the server upserts by AW id + aggregates per app/day — this only
  affects per-app attribution, never billed hours.
- The checkpoint is still computed from the **original** events, so a merged span
  can't rewind or double-advance it.
- Frozen source events are never mutated; the activity/fraud analyzer still sees
  the raw events.

## Tests
- 8 unit tests (`tests/test_window_flicker_coalesce.py`): flicker run merges &
  clears the filter, different apps not merged, already-passing events untouched,
  flicks absorbed into an adjacent same-app event, gap breaks the run, still-brief
  runs stay suppressed, source not mutated.
- Full suite: **847 passed**.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -1,4 +1,4 @@
 """BetterFlow - ActivityWatch to BetterFlow sync companion app."""
 
-__version__ = "1.5.88"
+__version__ = "1.5.89"
 __author__ = "BetterQA"
--- a/src/sync/sync_engine.py
+++ b/src/sync/sync_engine.py
@@ -1220,12 +1220,24 @@ class SyncEngine:
         new bucket-processing path can't silently classify against an empty
         context (this is the window-classification entry point).
         """
-        # Feed window events to activity analyzer for window change detection
+        # Feed window events to activity analyzer for window change detection.
+        # The analyzer sees the RAW events (unchanged) — coalescing below only
+        # affects what we upload, not window-change/fraud detection.
         if bucket_type in (BUCKET_TYPE_WINDOW, BUCKET_TYPE_WINDOW_ALT):
             self._activity_analyzer.add_window_events(events)
 
+        # Coalesce same-app title flicker so a run of sub-minimum window events
+        # isn't dropped wholesale by the flicker filter (per-app attribution
+        # loss). The checkpoint below is still computed from the ORIGINAL events,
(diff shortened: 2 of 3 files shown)


tudorsss:
Closing as superseded.

`main` shipped a more complete fix for this the day after this PR was opened: `_coalesce_window_flickers` in `src/sync/sync_engine.py:2112`, merged as #118 (`ef3c0f2`, 2026-07-02).

That implementation goes further than this draft — it merges on exact same-window (app + title + url) with a 2s gap tolerance, and handles checkpoint-based dedup so a flicker run spanning multiple sync cycles cannot double-send. It is covered by 11 tests in `tests/test_window_flicker_coalesce.py`.

This branch is now CONFLICTING against that work. Closing rather than rebasing, since the problem is solved.

PR #131, opened 2026-07-13
fix(queue): surface + bounded-replay dead-letter; close watchdog stacked-chain gap

tudorsss:
Two audit findings.

## P3 (primary): dead-letter table written but never replayed or surfaced

PR #129 made remove_failed MOVE exhausted events into dead_letter_events (preserved for inspection/replay), but nothing in production surfaced the table's size or re-enqueued its rows. A silently growing dead-letter is real lost activity no one can see.

- **Visibility**: SyncCoordinator._build_health_telemetry now reports dead_letter_count on every heartbeat, so a growing table is visible to ops.
- **Bounded replay**: new OfflineQueue.requeue_storable_dead_letter() resurrects rows that are storable AGAIN into the live queue. Same storable/retention classification as evict_unstorable (bucket present AND timestamp within the retention window); genuinely-unstorable rows (no bucket / stale) are left behind. Conservative: bounded batch, MOVE-not-copy (a row can't be resurrected twice, no double-send), retry_count reset, and a 30-min cooldown so a just-dropped definitive-4xx batch is not instantly re-resurrected into the same rejection (which would re-block the queue head and thrash drop/resurrect/drop). Wired into _process_queue past the backoff gate, so nothing is resurrected into an outage.

## P4 (secondary): narrow watchdog residual

The budget gate is checked before starting a chain, but the send loop's first group ran unconditionally. A session-start chain (~94s at elapsed 0) followed by the unconditional first-send chain (~94s) could reach ~188s, over the 150s _do_sync watchdog. Fix: gate the first send group on the same in-cycle network budget as the rest, so at most one chain survives the cut (budget 50s + one chain ~94s = ~144s). Forward progress across cycles stays with the durable OfflineQueue.

## Tests (TDD, verified fail pre-fix / pass post-fix)

- test_queue: replay resurrects only-storable, bounded/no-double-send, cooldown holds fresh drops then resurrects after it elapses.
- test_dead_letter_telemetry: dead_letter_count in telemetry (incl. 0); _process_queue drives the replay; the stale row stays behind.
- test_sync_send_budget: over-budget queues ALL groups incl. the first; a session-start chain plus a send chain do not stack past the watchdog.
- test_sync_watchdog_budget: two stacked chains would exceed the deadline invariant.
- Regression guarded: test_queue_no_drop_on_outage (head-of-line-block drop still holds; the cooldown is what keeps the replay from undoing it).

Full suite: 967 passed. No server / edge-function / migration component; reaches users on the next tagged desktop release.

Held at PR (not auto-merged): concurrent sessions were active on this worktree during the change, so the multi-session hard rail suspends the ecosystem auto-merge default.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -132,11 +132,13 @@ not the person using it:
   surface. A VM, a container or a locked-down Linux box legitimately reports
   `null`.
 - `timezone`, `agent_version`, and the tracker-health telemetry (restart counts,
-  event ages, sync staleness, and the two capture-dead flags
+  event ages, sync staleness, `dead_letter_count` — how many captured events
+  this device failed to DELIVER and is holding locally, a bare integer carrying
+  no event, bucket, timestamp or title — and the two capture-dead flags
   `tracker_download_failed` / `managed_components_unavailable`, which say the
   tracker binaries could not be installed and that this process has no managed
   watchers of its own). All of it describes whether the machine is capable of
-  recording, never what was recorded.
+  recording and delivering, never what was recorded.
 
 The serial is shown back to the user in the tray under **Diagnostics > Device
 serial**, next to the Privacy Policy link, and clicking it copies the value.
--- a/src/disclosure_baseline.py
+++ b/src/disclosure_baseline.py
@@ -535,6 +535,21 @@ HEARTBEAT_HEALTH_KEYS: tuple[str, ...] = (
     # fact about whether a watcher on this machine is working, never about what
     # the person did.
     "window_tracker_blind",
+    # How many events this device captured but failed to DELIVER — they
+    # exhausted their retries and were preserved locally rather than deleted.
+    # A single integer, and only ever a count: it carries no event, no bucket,
+    # no timestamp, no window title, and nothing that distinguishes one
+    # person's activity from another's. It says "this machine is holding N
+    # undelivered records", which is the delivery-health counterpart of
+    # consecutive_sync_failures and sync_stale_seconds, both already here.
+    #
+    # It reports a delivery FAILURE, so it can only ever describe the agent
+    # doing LESS than the notice says, never more. It was already computed on
+    # every heartbeat and filtered out at this gate, which is why no device has
+    # ever reported a backlog. Same reading as tracker_download_failed above:
+    # if the category is ever contested, the safe answer is to keep the field
+    # and update the notice text, not to drop it and go back to blind.
+    "dead_letter_count",
 )
 
 #: The machine's hostname is read once and embedded in every `bucket_id`, so
--- a/src/main.py
+++ b/src/main.py
@@ -1777,6 +1777,16 @@ class SyncCoordinator:
         last_ok = self._last_successful_sync
         if last_ok is not None:
             telemetry["sync_stale_seconds"] = int(time.monotonic() - last_ok)
+        # Surface the dead-letter table size so a growing backlog of preserved
+        # (but never-delivered) events is visible to ops. Nothing else reported
+        # it, so a silently climbing dead_letter_events — real lost activity —
+        # was invisible. Reported unconditionally (even 0) as an explicit health
+        # signal. Best-effort: a queue that can't answer must never block the
+        # heartbeat.
+        try:
+            telemetry["dead_letter_count"] = self.queue.dead_letter_count()
+        except Exception as e:  # noqa: BLE001
+            logger.debug("dead_letter_count telemetry failed: %s", e)
         try:
             telemetry.update(self.aw_manager.health_snapshot())
         except Exception as e:  # noqa: BLE001
--- a/src/sync/bf_client.py
+++ b/src/sync/bf_client.py
@@ -487,6 +487,16 @@ class BetterFlowClient(BaseApiClient):
         # see a blind idle tracker but never a blind window tracker. Same
         # category as its sibling — the agent reporting on its own watchers.
         "window_tracker_blind",
+        # Fourth field of the same shape. Size of the local dead-letter table:
+        # events that exhausted their retries and were PRESERVED rather than
+        # deleted. A climbing value is real, captured, never-delivered billable
+        # activity sitting on one laptop. main._build_health_telemetry has
+        # always computed it (reported unconditionally, even at 0) and this
+        # allowlist dropped it — so no device has ever reported a backlog, and
+        # the "a growing backlog is visible to ops" requirement the whole
+        # preserve-don't-delete design exists for was never actually met.
+        # int, always present: 0 is an explicit healthy signal, not an omission.
+        "dead_letter_count",
     )
 
     def heartbeat(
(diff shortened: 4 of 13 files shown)


tudorsss:
Rebased onto current `main` (30f1add) and fixed a defect the rebase exposed.

**The rebase itself was clean — zero conflicts — which is what made this worth writing up.** Once rebased, two of this PR's own tests failed:

```
test_requeue_storable_dead_letter_resurrects_only_storable         assert 0 == 1
test_requeue_storable_dead_letter_bounded_and_no_double_resurrect  assert 0 == 2
```

`requeued == 0` reads as the replay being broken. It wasn't — the fixture was.

**Mechanism.** `remove_failed` was the only one of the three dead-letter functions without an injectable `now`; `evict_unstorable` and `requeue_storable_dead_letter` both have one. So `dropped_at` was stamped from the real clock while the tests injected `now=2026-07-13` into the reader. The replay selects `WHERE dropped_at <= cooldown_cutoff`, so the moment wall-clock passed 2026-07-13T12:00Z every dead-lettered row sorted *after* the cutoff and the query returned nothing.

These tests broke on 2026-07-13 — the day this PR was opened — and could never have passed again. The feature is fine.

**Fix** (`53941e7`): completed the injection rather than moving the fixture date, so fixture and every clock read move together. That is the pattern the other two functions already use.

**Verification**
- Fails with the injection removed (both tests, `requeued == 0`); passes with it.
- Fixture year swept to **2020** and **2099** — both green, so wall-clock no longer participates in the outcome.
- Full suite **1499 passed**, exit 0. `main`'s baseline is 1490, so all 9 tests this PR adds now pass.

**One thing for a reviewer to confirm before merge.** #112 proposed deleting `_STUCK_HEAD_CEILING`, which is the mechanism that feeds `dead_letter_events` and therefore what this replay consumes. I closed #112 as superseded by #129, so that hazard is resolved — but it is worth a second pair of eyes that the ceiling remains the intended producer here.

Not merging: at +1373 this warrants the pre-merge review gate, not an auto-merge.

tudorsss:
## Pre-merge audit gate — complete, but BLOCKED on a disclosure review

Two independent reviewers (correctness lens, design lens) ran in parallel on the branch diff. **Both independently found the same Critical**, which is the strongest signal this process produces.

### Findings, all evidenced

| # | Finding | Status |
|---|---|---|
| C1 | Replay re-derived storability inline, dropping the duration bound | fixed `3cc6494` |
| C2 | `dead_letter_count` absent from the egress allowlist — feature inert | fixed `ce6d5d0` |
| I1 | Send-budget gate could freeze delivery permanently | fixed `5fcb189` |
| I2 | "can never be resurrected twice" was a false invariant | fixed `69d0897` |
| I3 | Replay bypassed `enqueue()`'s capacity eviction | fixed `fe3b3eb` |
| I4 | `stale_after_days=7` duplicated four times | fixed `2510d2c` |

**C1 was the serious one.** `queue.py` checked bucket + timestamp; `is_event_storable` checks bucket + timestamp + duration. So a 90000s weekend-suspend span, correctly dead-lettered by `evict_unstorable` as poison, would be resurrected ~30 min later — *after* that cycle's eviction gate — batched with healthy events, 4xx the whole batch, and `increment_retry` would drag its healthy neighbours to the drop ceiling. Real lost billable time, repeating every 30 minutes for the full 7-day retention window. The existing test covered no-bucket and stale but never an over-long duration, which is why it shipped.

**I1 was the subtle one.** With `/session/start` degraded while `/events/batch` stays healthy, `need_session` is true every cycle and burns the shared 50s budget before the send loop — so the cycle sends nothing *and* drains nothing, forever, with a green heartbeat. The old code still delivered the first bucket group. Fixed with a starvation floor at 3 cycles.

### Scope note

The fixer also corrected three sibling atomicity bugs in `remove_failed`, `evict_unstorable` and `enqueue`, found by grepping I2's class rather than fixing only the reported instance. The scoped re-review judged all three safe, with one to watch: `evict_unstorable` now parses every queued row while holding the RESERVED write lock, so the 5s default busy timeout became correctness-adjacent. Accepted because all queue writers run on the sync thread, and the pre-fix outcome under the same overlap was a silent duplicate dead-letter row — double-billed time, strictly worse than a loud retried error.

### Verification

- Full suite **1513 passed**, 0 failed. Branch baseline before the fixes was 1499; `main` is 1490.
- Every fix has a witnessed RED — the test was watched failing before the fix and passing after.
- A retention guard that was vacuous (an `is` comparison, defeated by small-int interning) was found and rewritten to compare declaration text; confirmed it goes red under mutation.

### BLOCKER — human disclosure review required

Adding `dead_letter_count` to `HEARTBEAT_HEALTH_KEYS` correctly tripped `test_heartbeat_egress_allowlist_still_holds`. That list is the enforced allowlist of what leaves an employee's machine, and `src/disclosure_baseline.py` states plainly: **CHANGES HERE REQUIRE A DISCLOSURE REVIEW: dpo@betterqa.co**.

The mechanical half is done — the key is declared in the baseline with its reasoning and CLAUDE.md is in step. The remaining half is a human review under Law 190/2018, and it is not a check an agent gets to tick.

Assessment for whoever reviews it: `dead_letter_count` is a count of undelivered events, so it reports whether the machine is syncing, not anything the person did — the same category as `tracker_download_failed`, already on the list. That is a recommendation, not a decision.

**Not merging until that review happens.** The guard worked exactly as designed.

### Deferred (logged, not fixed)

1. Forced drain makes >150s cycles routine during a session-start outage, so `_DO_SYNC_DEADLINE` mints a report roughly 1 cycle in 3 — real alert noise.
2. `_consecutive_undelivered_cycles` neither increments nor resets across offline/empty stretches, so a stale count of 2 can force a drain on the first starved cycle after reconnect.
3. The `_cursor` docstring's "the contention window shrinks rather than grows" is inaccurate for `evict_unstorable`.
4. `sqlite3.connect` sets no explicit busy timeout; the 5s default is now correctness-adjacent.
5. The retention guard's line filter could false-fail on a future docstring line, and a false-failing guard is one someone loosens.

tudorsss:
## Disclosure review signed off — blocker cleared

The `dead_letter_count` egress addition has been through the disclosure review required by `src/disclosure_baseline.py`. Confirmed complete by the repo owner (Tudor Brad) on 2026-07-29; recorded here so the approval lives with the change rather than in a chat log.

Scope of what was approved: one new heartbeat field, `dead_letter_count`, an integer count of events currently sitting in the local dead-letter table. It reports whether the machine is delivering, not what the person did — the same category as `tracker_download_failed`, already on the allowlist. Declared in `HEARTBEAT_HEALTH_KEYS`, `src/disclosure_baseline.py` and CLAUDE.md's device-identifier section, which `test_disclosure_baseline.py` cross-checks for parity.

Merging. `main` is unchanged at `30f1add` since the gate ran and the branch head is still `5fcb189`, so the 1513-passing verification covers exactly the tree being merged.

PR #144, opened 2026-07-19
fix: audit remediation (13 commits, 2026-07-18)

BetterQAFTW:
Remediation pass from the 2026-07-18 audit: 13 commits off `main`.

Each commit is self-contained and explains the failure it fixes in its own message; the log below is the index.

## Commits

- fix(keychain): catch ValueError from credential validation
- fix(queue): guard corrupt created_at in QueuedEvent.from_row
- fix(updater): re-check host after download redirect
- fix(update-handler): log flush/shutdown failures during update exit
- fix(update-handler): guard coordinator.stop() during update exit
- fix(update-handler): log unparseable staged version instead of silent pass
- fix(aw-manager): re-check host after tracker download redirect
- fix(aw-manager): verify tracker archives against pinned SHA-256
- fix(aw-manager): surface a fail-closed tracker download
- refactor(url-safety): share one post-redirect host re-check
- fix(aw-manager): report download failure to ops before toasting
- fix(aw-manager): back off the tracker download, not just its toast
- fix(aw-manager): read tracker_download_failed under the lifecycle lock

---

https://claude.ai/code/session_01TcSd2noNuNxUtTYp9xArv7

### DIFF
--- a/.github/workflows/verify-tracker-pins.yml
+++ b/.github/workflows/verify-tracker-pins.yml
@@ -1,0 +1,47 @@
+name: Verify Tracker Asset Pins
+
+# The RELEASE_SHA256 pins in src/aw_manager.py are hand-computed literals, and
+# tests/test_aw_manager_asset_integrity.py only cross-checks them against a
+# second copy of the same literals — so a wrong value passes the unit suite and
+# fails closed on the fleet instead (no trackers installed, zero capture).
+# This job is the independent check: it fetches the real upstream archives and
+# asserts their digests equal the pins. It runs nightly so upstream mutating an
+# asset under a pinned tag is caught here rather than on user machines.
+
+on:
+  schedule:
+    - cron: "17 4 * * *"
+  workflow_dispatch:
+  push:
+    paths:
+      - "src/aw_manager.py"
+      - "tests/test_aw_manager_asset_integrity.py"
+      - "scripts/verify_tracker_pins.py"
+      - ".github/workflows/verify-tracker-pins.yml"
+
+# Read-only: this job downloads public upstream release assets and compares
+# hashes. It never writes to the repo or touches releases.
+permissions:
+  contents: read
+
+jobs:
+  verify-pins:
+    runs-on: ubuntu-latest
+    # The per-read socket timeout in the script does not bound total runtime: a
+    # slow-drip upstream can keep three ~100 MB fetches alive for hours on the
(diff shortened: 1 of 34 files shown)


BetterQAFTW:
## 2026-07-20 audit pass — 10 commits added

Branch now 25 commits (plus 2 landed after the auditor finished: `82236c9`, `ca12379`). Report: `RULES-AUDIT/betterflow-sync-2026-07-20.md`

### Two CRITICALs, both fixed

**1. The consent screen was lying.** First-run consent said *"Window titles are hashed on your device"* and the README claimed SHA-256 hashing. **There is no client-side hashing in the agent — titles are sent raw.** That screen is the designated consent boundary on Windows/Linux where there's no OS permission gate, so users were consenting on a false statement about what leaves their machine.

**2. A permanently dead input tap that reported healthy.** macOS disables a `CGEventTap` for two reasons; the code handled only the timeout constant. A tap disabled by user input was silently dropped — counters frozen at 0 forever while `is_running` stayed True, which downstream reads as *"user is idle"* rather than *"we're blind."* That is the phantom-idle signature that bills real work as idle, and it's a distinct mechanism from both the two-readers bug and the TCC-blindness one.

Three of the five findings carried over from 2026-07-18 were already fixed on the branch; re-verified rather than re-reported.

**Tests:** 1245 → **1249 passing, 0 failures** (10 added). Green at every commit. 0 tracked secrets.

### Still open — needs a decision, not a commit

- **`self_updater.py:418` (HIGH).** Windows/Linux self-update has zero authenticity verification while macOS has full codesign pinning, with `auto_install_updates` defaulting True. This has now survived **three audit passes** because it needs release infrastructure, not a code edit: a Windows publisher cert thumbprint that doesn't exist in the repo, and a signed Linux hash manifest the pipeline doesn't publish. Deliberately did **not** port just the downgrade check — without authenticity it defends against nothing while looking like mitigation.
- **`permissions.py:321` (HIGH).** Shows a native admin-password prompt to write `TCC.db`, which SIP has blocked since macOS 10.15. It always fails and returns `True` anyway. The working replacement already exists; deleting a user-facing prompt is a product call.
- Three further HIGH items (`is_running` semantics, a break-manager TOCTOU behind an explicit lock-ordering comment, an input-drain reset that tests document as deliberate) are written up with reasoning rather than half-fixed on the capture hot path.

The untracked `tools/fraud_sim/` predates this run and was left alone.

PR #146, opened 2026-07-20
fix(queue): sanitize project ids before draining status spans

BetterQAFTW:
## Summary
- normalize project IDs before stamping sync events so only backend-safe positive integer IDs are sent
- sanitize already-queued events on startup so invalid project_id values do not keep poisoning queue drains
- bump app version to 1.5.106 for the beta build

## Tests
- pytest tests/test_project_stamp_single_source.py tests/test_queue_storability_backfill.py tests/test_sync_engine.py::TestFailedStatusSpanIsStorable tests/test_sync_engine.py::TestStatusSpanEvents tests/test_queue_no_drop_on_outage.py tests/test_queue_mixed_batch_poison.py tests/test_release_version.py -q
- pytest -q

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -1,4 +1,4 @@
 """BetterFlow - ActivityWatch to BetterFlow sync companion app."""
 
-__version__ = "1.5.105"
+__version__ = "1.5.106"
 __author__ = "BetterQA"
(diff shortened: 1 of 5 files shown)


tudorsss:
Closing as already landed.

This exact change is on `main` as commit `82236c9` (2026-07-20) — same `sanitize_project_ids()`, same five files. GitHub reports this PR as CONFLICTING because it is trying to re-apply a change that is already there under a different commit hash.

No action needed; the fix is live.

PR #147, opened 2026-07-22
Windows input rollout: make the default-ON actually reach the fleet, and stop Private Time keystrokes being billed

BetterQAFTW:
## Why

Two Windows devices (Sachi, Claudia — both 1.5.116) show **`input sensor missing`, 0 keys/min, 0 clicks/min**.

Windows ships **no input sensor at all**. `BF_WATCHERS = ["bf-window-tracker", "bf-idle-tracker"]` — no input tracker is bundled, and `main.py` states plainly that "On Windows window_watcher/input_watcher are None". The only Windows input path is the in-process `InputSource` (ctypes `WH_KEYBOARD_LL` / `WH_MOUSE_LL`), which ships dormant and has never been enabled on a single device.

The commit that flips it on for Windows was written on 2026-07-22 and has been sitting in this PR unmerged since.

## What was actually broken

Auditing the rollout before merging found that **merging it as it stood would have fixed nothing**, plus a set of defects that would have made a blocked hook invisible — the exact condition we cannot currently distinguish from an idle employee.

1. **A persisted config outranks the platform default.** `update_from_server()` ends with `save()`, and `save()` wrote `sync.in_process_input`, so every Windows agent has `false` on disk from the dormant builds. `_from_dict` rebuilt `SyncSettings` from it, so the new default would have reached **fresh installs only**. Fixed the way `foreground_activity.enabled` already is: dropped on load, never persisted, server still authoritative.
2. **A refused hook reported a working sensor.** `available()` answered "is this Windows?", not "did `SetWindowsHookEx` succeed", and the source latched on that probe.
3. Any other escape from the hook thread left no verdict at all.
4. The availability latch outlived the backend it vouched for.
5. `GetMessageW` returns **-1** on error; the `!= 0` pump accepted it as a message and dispatched an uninitialised `MSG` in a tight spin.
6. Config blocks were **aliased, not copied**, and a non-mapping block raised inside `load()`'s `except` — discarding the **whole** config and silently falling back to defaults.
7-10. `Thread.start()` raising, a dead thread keeping `_start_ok`, a macOS watcher leaked **every 60s forever** on a denied grant, and an unguarded poll loop.
11. Both backends reported available after a clean `stop()`.
12-13. A no-op monkeypatch, and a test reading the runner's platform.

Plus, separately: **`test_in_process_input_server_flag_round_trip` would have gone red on the `windows-latest` leg of the release-tag matrix** — green on the ubuntu PR job, so it would have surfaced while cutting the release.

## The privacy bug this armed

`_advance_checkpoints_to_now` skips the paused window for the AFK and window streams but never got the input one, and never cleared the counters. **Keystrokes typed during Private Time drained into the first span after resume and were billed** — measured pre-fix at 500 private presses + 7 real ones arriving as a single event of **507**.

Dormant only because `in_process_input` is off fleet-wide. **The Windows default-ON in this PR is what arms it, on every Windows device at once.** Counts are now discarded, on the source *and* in the backend's own buffer (macOS folds the wrapped watcher's counts in a second later, and a dead poll thread could resurrect a whole window).

## Also in here

- Working hours judged in the device's **live** timezone, self-healing (the original reason for this PR — Carmen's LA-anchored 0h day).
- Queue backoff cleared after a confirmed delivery.
- Auth tolerance widened to a sustained window, so a transient 401 can't log a device out.

## Verification

- **1447 pass, exit 0.** Every fix has a test that fails pre-fix; every guard is mutation-tested (disable it, the suite goes red).
- The `-1` pump test is deliberately **bounded** — the pre-fix loop *hangs* rather than fails, and a guard that hangs CI is worse than one that goes red.
- Merge conflict resolution re-added `schedule_timezone_mismatch` to what `main` turned into the enforced heartbeat egress allowlist; taking `main`'s side verbatim compiled, passed, and would have silently killed that telemetry. It is now declared in `disclosure_baseline.py`.

## ⚠️ Not verified, and it is the part that matters

**The Windows hook path in this branch has never executed.** Every test driving it uses a fake `user32` on a Mac — `SetWindowsHookEx` has not been called once. This is precisely the surface `browser-platform-verification.md` says CI cannot see, and four of the defects above lived in it.

**Required before any fleet rollout:** install on ONE Windows machine, confirm `keys/min` goes non-zero and the `input sensor missing` chip clears. Do not roll wider on green CI.

Claudia's `0m` / 6-hour-stale sync is a **separate, undiagnosed** issue. `paused_by_network` early-returns both the heartbeat floor and the 60s tick and is cleared only by an OS network-online event, wake, or a user pause toggle — no reachability self-heal, though `_set_sync_failure_state` already has the probe. That is the only shape in this codebase that freezes Last Sync and Last Seen at the same instant. **Hypothesis, not a diagnosis** — it needs her agent log, and it is not fixed here.


BetterQAFTW:
## On-device verification (macOS, this build)

Installed the signed `BetterFlow-macOS-arm64.dmg` from run [30021556120](https://github.com/Better-Quality-Assurance/betterflow-sync/actions/runs/30021556120) (all four platforms green) onto a real Mac running 1.5.116.

**Signature / TCC continuity** — the risk was that a differently-signed bundle silently drops the Accessibility and Input Monitoring grants while the checkboxes still read enabled:
```
app inside DMG: Developer ID Application: Better Quality Assurance SRL (87NVC57J44)
spctl:          accepted — source=Notarized Developer ID
Identifier:     co.betterqa.betterflow      (identical to the installed 1.5.116)
post-install:   Input Monitoring diagnostics: hid_access=0 granted=True
```
Grants survived. Input counting is live, not merely started:
```
15:55:26  dur=10.0s  presses=2   clicks=1  scrolls=139
15:55:06  dur=10.0s  presses=40  clicks=0  scrolls=0
```

**The headline fix, on real hardware rather than a probe.** This machine's `config.json` carried the stale value — the same line every Windows agent has, rewritten every 30 minutes by `update_from_server`'s `save()`:
```
before (1.5.116):  "in_process_input": false   present
after  (1.5.117):  key absent
siblings batch_size / in_process_afk / in_process_window / interval_seconds intact
```
Surgical: one key dropped, every sibling preserved. On Windows that load now yields the platform default `True` instead of the stale `false`, which is the whole point.

**One more bug fell out of going to look for the evidence.** I expected a log line confirming the drop; there wasn't one. `Config.load()` ran one line *before* `setup_logging()`, so every load-time message — including the pre-existing `foreground_activity.enabled` one — went to a logger with no handler. Fixed in 4d588b7 with an AST callsite guard, because the obvious test is a phantom (`caplog` attaches its own handler and passes in either order).

## Still unverified — the part that matters

**The Windows hook path has never executed.** Every test driving it uses a fake `user32`; `SetWindowsHookEx` has not been called once. Installing on macOS does not touch it.

`BetterFlow-Windows` from the same run needs a real Windows machine. Pass condition: **keys/min goes non-zero and the `input sensor missing` chip clears.** Do not roll wider on green CI — four of the defects fixed here lived in exactly that path.

PR #176, opened 2026-08-03
fix(sync): cross a multi-day empty backlog gap in one cycle, not one 2h window per cycle

BetterQAFTW:
## The bug

After a weekend, the window/input bucket checkpoints sit ~62h behind. The empty-window gap-jump in `_fetch_bucket_events` advanced only ONE `_BACKLOG_WINDOW` (2h) per 60s sync cycle, so crossing the weekend took ~31 cycles = **~31 minutes** during which the morning's window/input events were captured locally but never uploaded. The AFK checkpoint isn't pinned, so billing kept flowing — but the server (correctly) fired the `window_ingest_stalled` / app-attribution-gap alert. This hits **every device with a weekend gap, every Monday**.

## Live evidence (device 14, 2026-08-03)

- Agent log: `1 sent` (the AFK event) every cycle 07:36–08:05 EAT, then a **139-event flush at 08:06:44** — exactly 30 min after start.
- Prod `agent_events`: the morning's window rows all have `created_at` 05:06:39–05:08:39 UTC while their `event_timestamp` starts at 04:34 UTC.
- First upload of the day carried a Jul-31 14:40 tail — the pinned checkpoint.

## The fix

Walk consecutive empty windows **inside** one `_fetch_bucket_events` call (each step is one local AW query; the checkpoint advances per step so an interrupted walk resumes), capped at `_EMPTY_GAP_MAX_WINDOWS_PER_CYCLE = 64` (~5.3 days/cycle) so an ancient checkpoint can't turn one cycle into thousands of fetches.

## Proof-of-failure handshake

- `test_multi_day_gap_crossed_in_single_cycle`: **failed pre-fix** (`batch` was `[]` — the single jump advanced only 2h into a 62h gap), passes post-fix.
- `test_empty_gap_walk_is_bounded_per_cycle`: **failed pre-fix** (cap didn't exist), passes post-fix; also asserts AW round-trips per cycle ≤ cap+1.
- The two pre-existing idle-gap-pin tests pass before and after (they pin correct behaviour, not this defect).
- Full suite: **1560 passed, 1 skipped** (CI runs pytest only; ruff findings on these files are pre-existing on main, verified via stash baseline).

https://claude.ai/code/session_01PutE4XEaZQu7YMBF9bUdnF

### DIFF
--- a/src/sync/sync_engine.py
+++ b/src/sync/sync_engine.py
@@ -215,6 +215,14 @@ class SyncEngine:
     # (~300 input/h), so the slice keeps the true oldest events.
     _BACKLOG_WINDOW = timedelta(hours=2)
     _BACKLOG_FETCH_LIMIT = 1000
+    # An EMPTY span (weekend, leave) is walked window-by-window INSIDE one
+    # cycle, not one window per cycle — at one 2h step per 60s cycle a 62h
+    # weekend takes ~31 min, during which the morning's window/input events sit
+    # captured-but-unsent and the server's 15-min window-ingest-stall alert
+    # fires (device 14, 2026-08-03). Each empty step is one cheap local AW
+    # query; the cap bounds a cycle when the checkpoint is ancient (64 × 2h =
+    # ~5.3 days per cycle, so any realistic gap clears in 1-2 cycles).
+    _EMPTY_GAP_MAX_WINDOWS_PER_CYCLE = 64
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


BetterQAFTW:
Closing as redundant: #175 (merged 4 min before this) fixes the identical defect with the same in-cycle empty-window walk, and is a superset — it adds a wall-clock budget (`_BACKLOG_WALK_BUDGET_SECONDS`) bounding the walk against a degraded local AW server (10s timeout × 48 probes × 4 buckets would blow the 150s sync watchdog), which this PR lacked. Its tests cover both of this PR's cases (single-cycle crossing, per-cycle bound) plus a mid-gap cluster stop. Independent diagnosis of the same live incident (device 14, 2026-08-03 Monday-morning stall) reaching the same mechanism is good corroboration; one implementation is the right end state.

PR #191, opened 2026-08-14
fix(updater): never offer a DMG the machine cannot execute, and stop a timeout pinning the wrong arch

tudorsss:
Two follow-ups to #185, found auditing it. Neither is a defect Martin introduced; both are gaps the fix left standing next to it.

## 1. The fallback could still hand a Mac a build it cannot run

`_find_platform_asset`'s fallback loop matched **any** macOS `.dmg`. So a release missing its arm64 asset served an Apple Silicon Mac the `x86_64` build — which, on a machine without Rosetta 2, installs a binary that dies `EBADARCH`/`ENOEXEC` and stops capture dead. That is the production fault already on record for Ardiel Plata's device (internal-tool2 #2298). The Intel direction is worse: there is no reverse Rosetta, so an arm64 DMG on a genuine Intel Mac has no recovery path at all.

The fallback **predates** #185. What #185 changed is that it became reachable — before it, a Rosetta install resolved `arch=x86_64` and always matched the Intel DMG in the first loop, so it never fell through. Resolving the hardware arch is precisely what lets the first loop miss.

Fixed at **both** addresses rather than the one that was reported: the trailing ZIP loop is shared with macOS and carries the identical hazard for a `BetterFlow-macOS-x86_64.zip`. One shared `_is_wrong_arch()` instead of the condition inlined twice.

An asset naming **no** architecture still qualifies (it may be universal); only one naming the *other* arch is refused. An undetermined arch — `platform.machine()` may return `""` — now rejects every arch-suffixed asset rather than matching all of them through the empty substring.

## 2. A timed-out probe was memoised for the process lifetime

`_read_proc_translated_cached` cached whatever the probe returned, including the `None` from an exception. One `sysctl` timeout therefore marked the machine untranslated for the whole session: the Diagnostics row states "Intel" about an Apple Silicon Mac, and the next update check re-selects the Intel DMG. #185's bug, silently back, with nothing to show for it.

Not theoretical — the warm-up fires at the busiest moment of launch, in a process that in the affected case is itself running through Rosetta, against a `timeout=2`.

The docstring justified caching failures with the Intel-Mac case, which is real and permanent: the key does not exist and `sysctl` exits **non-zero**. It then applied that rule to the exception path, which is neither. The line is exit code vs exception — "asked and answered no" vs "never got an answer". Retries are capped at 2 so a sandbox blocking `sysctl` outright still cannot fork on every menu rebuild.

## 3. Two tests that were not testing what they said

- `test_probe_runs_at_most_once_per_process` modelled a genuine Intel Mac by raising `OSError`. A real Intel Mac does not raise; it exits non-zero. Repointed at `returncode=1`, so it pins the conclusive path it always meant to.
- Every assertion about `send_notification` was a **negative** one (`assert_not_called` on a broken probe), and the ordering tests read `run()`'s source for the *call site*, never the body. Emptying the `send_notification` block left the entire suite green — the user-facing half of #185 was witnessed by nothing, on a path that runs on no CI leg and essentially no developer machine. Added the positive control.

## Verification

Proof of failure for the fallback fix — both new tests fail against the pre-fix tree:

```
assert 'https://x/BetterFlow-macOS-x86_64.dmg' is None
assert 'https://x/BetterFlow-macOS-arm64.dmg' is None
```

while the other 6 in that file pass unchanged, so the suite isolates the defect rather than describing the new code.

Mutation matrix — four mutants, each killed by a **distinct named** test, plus a clean control run:

| mutant | test that reddened |
|---|---|
| neuter `_is_wrong_arch` | `test_apple_silicon_is_never_handed_the_intel_dmg_as_a_fallback` + `test_an_intel_mac_is_never_handed_the_arm_dmg_as_a_fallback` |
| restore the unconditional memo | `test_a_transient_probe_failure_is_not_memoised` |
| drop the attempt cap only | `test_a_permanently_broken_sysctl_still_stops_forking` |
| empty `send_notification`, keep the guard and its logging | `test_a_translated_mac_actually_gets_the_notification` |

The last one is the consequence mutant rather than the condition mutant — the `logger.warning` still fired in the captured log, so the guard was intact and only its effect was gone. Each mutation was confirmed applied with `cmp` before the run.

- Full suite: **1676 passed, 1 skipped, exit 0** (up from 1656 at #185, so the run reached the new tests).
- `ruff check` on all five changed files: All checks passed.
- CI's test job is `pytest tests/` on ubuntu-latest / Python 3.11. Every test here injects `system="Darwin"` and patches `subprocess.run`, so none of them consults the host — the pattern #186 established for exactly this reason.

## Not included

This only reaches users when a release is cut. No release bump here — that is a separate call.

### DIFF
--- a/src/machine_arch.py
+++ b/src/machine_arch.py
@@ -19,6 +19,7 @@ Kept dependency-free and fully injectable (``machine`` / ``translated`` /
 same spirit as ``release_version.py``.
 """
 
+import errno
 import logging
 import platform
 import subprocess
@@ -63,16 +64,27 @@ class ProbeResult(NamedTuple):
     conclusive: bool
 
 
-# Only a TIMEOUT is worth retrying. A missing sysctl or a sandbox denial will not
-# resolve itself mid-session, so those are conclusive and get memoised like any
-# other answer. Classifying them as transient is what broke the Windows CI leg
-# in review: no sysctl there, so FileNotFoundError made the memo never engage and
-# the fork-counting test saw a second fork.
-_RETRY_AFTER_SECONDS = 60.0
+# CONGESTION, in every spelling it reaches us in. A timeout is the obvious one,
+# but the same overloaded machine also refuses a fork outright (EAGAIN, when the
+# per-user process table is exhausted) or has the child killed under memory
+# pressure (jetsam/OOM, which surfaces as a negative returncode). All three say
+# "the machine was too busy", none says anything about its hardware, so none may
+# be memoised.
+#
+# The distinction is NOT exception-vs-exit-code, which was this module's second
+# wrong cut at it: FileNotFoundError (no sysctl — every Windows host) and
+# PermissionError (a sandbox denial) are permanent for this process and MUST be
+# memoised, or the memo never engages and menu rebuilds fork forever.
+_TRANSIENT_ERRNOS = frozenset({errno.EAGAIN, errno.ENOMEM})
 
-# Even a persistently timing-out sysctl has to settle, or a long session keeps
-# paying a fork per minute for an answer that is not coming.
-_MAX_PROBE_ATTEMPTS = 3
+# Retries back off, and that is the point rather than politeness. The congestion
+# that loses the first probe is a boot storm — Spotlight reindexing after an OS
+# update, an MDM scan, Time Machine — and those last minutes, not seconds. A flat
+# interval with a small cap spends the whole budget inside the storm and settles
+# on the wrong answer just as the machine calms down, which is the same defect as
+# a count-based budget wearing a clock. This schedule keeps trying past the hour
+# for a total of four forks.
+_RETRY_BACKOFF_SECONDS = (60.0, 300.0, 1800.0)
 
 
 def _monotonic() -> float:
@@ -89,7 +101,7 @@ def _read_proc_translated_cached() -> Optional[str]:
     not exist and never will, so a probe that retried would spawn a subprocess
     on every menu rebuild forever.
 
-    A TIMEOUT is a different thing wearing the same return value, and memoising
+    CONGESTION is a different thing wearing the same return value, and memoising
     it silently reinstates the bug this module exists to fix: the machine reads
     as untranslated for the rest of the session, the Diagnostics row states
     "Intel" about an Apple Silicon Mac, and the next update check re-selects the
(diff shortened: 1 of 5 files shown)


tudorsss:
## Round 2 — pre-merge audit gate findings, fixed

Two reviewer lenses ran on the branch diff. Both independently found the same **Critical in my own round-1 fix**, one with a measured PATH-varied differential. Rebased onto `main` (now carrying the 1.5.123 release) and re-verified.

### Critical: round 1 would have broken the tagged release build

Classifying every exception as transient made `FileNotFoundError` transient too — and on Windows there is no `sysctl` at all. The memo never engaged, so `test_the_memo_makes_every_later_menu_rebuild_free` saw a second fork.

That test is green on every Mac and on the ubuntu PR job (`sysctl` resolves, exits non-zero, conclusive). It fails only on the **windows-latest leg of the tag-gated build**, and `release:` declares `needs: build` — so v1.5.124 would have published no artifacts. Invisible on the PR by construction.

Differential, same machine, only `PATH` varied:

```
round 1, sysctl absent:  1 failed, 6 passed   <- FAILED test_the_memo_makes_every_later_menu_rebuild_free
round 2, sysctl absent:  7 passed
round 2, sysctl present: 7 passed
```

Control: `which sysctl` in that shell confirms it is genuinely unreachable, so the probe measured what it claims.

Fix: a missing binary or a denied call **is** an answer — `sysctl` will not appear mid-session and a sandbox will not lift. Only `TimeoutExpired` is "never got an answer", and only it is retried.

### Important: the retry did not survive the condition its own rationale describes

`_probe_attempts` counted over the process lifetime with no time dimension. But the two probes of a launch are consecutive statements — `main.run` warms the memo, `tray.run_blocking` builds the menu milliseconds later. So a count-based budget was spent entirely **inside** the congested window that caused the timeout: the retry landed in the same conditions as the failure, the second timeout memoised the wrong answer anyway, and it cost an extra fork under `_menu_lock` for nothing.

Now gated on **elapsed time**, so the retry lands somewhere the machine is no longer busy, and rapid menu rebuilds stay fork-free in the meantime. Clock injected via `ma._monotonic`, so no test sleeps and none assumes the clock's *magnitude* (a fresh runner's uptime is ~95s).

### Design: the module-level flag is gone

`_read_proc_translated` now returns a `ProbeResult` NamedTuple, so the classification cannot be read out of order and the producer has no side effect its docstring had to warn about. The cited constraint — tests patch it as a zero-arg callable — cost exactly one fixture line, as the reviewer measured.

### Also fixed

- **`_is_wrong_arch` is scoped to Darwin.** Its unknown-arch branch was platform-blind, and the docstring's claim that Linux tokens "are not keys" was plainly false — `x86_64` *is* a key. Linux only escaped through the caller's early return, which is control flow, not a property of the rule.
- **The refusal comment was contradicted by its own fixture.** A machine resolving to `arm64` *because* `proc_translated == "1"` is running under Rosetta, so the Intel DMG would execute. Refusing it is a product choice (don't re-pin to Rosetta), not an inability. The EBADARCH case is the **native** arm64 one. Both are now named separately, and the note that a refusal still leaves the user a manual download is recorded.
- **Three mechanisms that mutation showed surviving now have tests**: the ZIP loop's guard, the unknown-arch rejection, and the `arch and` guard.
- **The retry-bound test asserts the property**, not just `ma._MAX_PROBE_ATTEMPTS` — comparing the module to itself stayed green if the bound grew to 50.
- The notification assertion is case-insensitive; a reword should not redden it.

### Verification

- **1680 passed, 1 skipped, exit 0** on the rebased branch (1676 at round 1).
- `ruff` clean on all five changed files. Repo-wide `ruff` reports **169 errors on this branch and 169 on `origin/main`** — pre-existing, unchanged.

### Note on release timing

**v1.5.123 shipped while this was in review**, so it carries the wrong-arch fallback. The fix reaches users at 1.5.124.

PR #213, opened 2026-08-24
Say why a Mac records nothing, and which build it is running

### LINKED ISSUE #184: Mac users can sit on the Intel build with no way to tell, and updates keep them there
## Business impact

A Mac user who picks the Intel installer on an Apple Silicon machine gets a working app, so nothing ever tells them they chose wrong. It runs under Rosetta 2 — slower, heavier on battery, and with known permission/TCC edge cases this codebase already documents. There is currently **no way for the user or for us to find out which build anyone is on**: the agent does not report its architecture, so the fleet view cannot answer "who is on Intel?".

Worse, it is self-perpetuating. The auto-updater picks its download by asking the OS what architecture it is on, and under Rosetta the OS answers "x86_64" — so an Intel install downloads the Intel build again on every single update, forever. Nobody drifts back to the right build on their own.

Reported by Lorenzo Muresan, whose team installed the Intel version on M1 laptops and asked for (a) the correct M1 link and (b) a way to see which version they are running.

## Steps to reproduce

1. On an Apple Silicon Mac, install `BetterFlow-macOS-x86_64.dmg` from the download page.
2. Launch it. It starts and works normally — no warning, no degraded-mode notice.
3. Try to determine from the app UI which build you are running. There is no such indicator.
4. Wait for an update, or trigger the update check. It fetches the x86_64 DMG again.
5. Check the backend fleet view for that device's architecture. It is not recorded.

To confirm the underlying mechanism on any Apple Silicon Mac (no BetterFlow needed):

```bash
python3 -c "import platform; print(platform.machine())"                 # arm64
arch -x86_64 /usr/bin/python3 -c "import platform; print(platform.machine())"   # x86_64
sysctl -n sysctl.proc_translated                                        # 0 native, 1 under Rosetta
```

To check which build a Mac currently has installed:

```bash
lipo -archs /Applications/BetterFlow.app/Contents/MacOS/BetterFlow       # arm64 or x86_64
```

## Expected result

- The app tells the user when it is running the Intel build on Apple Silicon, and offers the correct one — the same thing Apple does for Xcode ("Support Ending for Intel-Based Apps … Learn how to update to an Apple silicon version").
- The updater installs the build matching the real hardware, so a wrong install self-corrects on the next update instead of persisting.
- The backend knows each device's architecture, so we can answer "who is on Intel?" without asking people to run terminal commands.

## Actual result

- No architecture indicator anywhere in the app.
- No architecture field reaching the backend.
- The updater re-selects the Intel build indefinitely on a Rosetta-translated install.
- The only way to answer the question today is a manual `lipo` on each laptop.

## Technical details

**1. Architecture is never collected.** `src/sync/bf_client.py`, `DeviceInfo.collect()` gathers `hostname`, `os_name`, `os_version` (`platform.release()` — the Darwin kernel version, not the arch), `agent_version`, `machine_id`. No architecture field, so the heartbeat and `agent_devices` cannot carry one.

**2. The updater is Rosetta-blind.** `src/update_checker.py:78` in `_find_platform_asset()`:

```python
arch = arch or platform.machine()
```

then matches `arch in name` against the release assets. Verified above: under Rosetta `platform.machine()` returns `x86_64` on Apple Silicon, so this matches `BetterFlow-macOS-x86_64.dmg` and reinstalls Intel every time. `platform.machine()` cannot distinguish a real Intel Mac from a translated process — both report `x86_64`. The only reliable discriminator is `sysctl.proc_translated`.

Suggested helper, used by all three sinks below:

```python
def true_machine_arch() -> str:
    """Real hardware arch, seeing through Rosetta 2 translation."""
    if platform.system() != "Darwin":
        return platform.machine()
    try:
        translated = subprocess.run(
            ["sysctl", "-n", "sysctl.proc_translated"],
            capture_output=True, text=True, timeout=2,
        ).stdout.strip()
    except Exception:
        return platform.machine()
    # proc_translated == "1" means this x86_64 process is running on arm64 hardware.
    return "arm64" if translated == "1" else platform.machine()
```

Note the key is absent entirely on Intel Macs (not `0`), so treat "missing" and `0` the same — do not branch on presence.

**Affected sinks (one root cause, three places it surfaces):**

| Sink | File | Change |
|---|---|---|
| Auto-updater picks wrong DMG | `src/update_checker.py:78` | use `true_machine_arch()` instead of `platform.machine()` |
| Backend cannot see who is on Intel | `src/sync/bf_client.py` `DeviceInfo.collect()` | add an `arch` field, send it on register/heartbeat; backend stores it on `agent_devices` and shows it in the fleet view |
| User cannot tell what they are running | app UI | show the arch in About/status, plus the notification below |

**3. Two notifications, per the Xcode model.**

- **Already installed (translated):** on startup, if `proc_translated == 1`, show a persistent, dismissible notice — "BetterFlow is running the Intel version on an Apple Silicon Mac. Switch to the Apple Silicon version for better performance and battery life." with a direct link to the arm64 DMG. Should not nag every launch; once per version is enough.
- **Newly installing:** the download page cannot reliably detect this (see below), so the first-run check in the installed app is the reliable catch. If the installer itself can run a preflight, warn there too.

**4. On detecting architecture from the download page** — asked separately, and the answer constrains the design:

- The `User-Agent` string on macOS reports `Intel Mac OS X` **even on Apple Silicon**, in both Safari and Chrome, deliberately frozen for compatibility. This is true of the HTTP header too, so the server cannot tell either. UA sniffing therefore fails in the worst possible direction: it would confidently label every M1 user as Intel and send them to the wrong build.
- `navigator.userAgentData.getHighEntropyValues(['architecture'])` does return `arm` on Apple Silicon, but it is Chromium-only — no Safari, no Firefox — and Safari is a large share of Mac traffic.
- The WebGL `WEBGL_debug_renderer_info` renderer string ("Apple M1/M2" vs "Intel Iris") covers more browsers but is fragile and increasingly masked.

So the page can *hint* — highlight or reorder the recommended button when a high-confidence signal is available — but must never silently decide, and both buttons must stay visible. (Marked **inferred**: these are well-established browser behaviours but were not re-verified on-device for this issue; a live probe of `navigator.userAgentData.getHighEntropyValues(['architecture'])` and the UA string on an M1 would confirm before anyone builds on it.)

**5. The durable fix that removes the whole class:** ship a **universal2 binary** — one DMG containing both architectures — so there is no choice to get wrong, no detection to get wrong, and no wrong-arch state to recover from. PyInstaller supports `target_arch='universal2'`, though it requires every bundled native wheel to be universal too, so it is a real piece of work rather than a flag flip. Worth costing before investing in the detection path, since it makes items 1-4 largely unnecessary.

## Environment

- Repo: `Better-Quality-Assurance/betterflow-sync`; `src/update_checker.py`, `src/sync/bf_client.py`
- Agent: v1.5.122 (latest release, 2026-08-10)
- Verified on: macOS, Apple Silicon (arm64), Darwin 25.5.0 — Rosetta behaviour reproduced directly with the commands above
- Existing Rosetta awareness already in the codebase: `src/ui/permissions.py:53,84,118` documents TCC false-negatives for "unsigned x86_64 binaries under Rosetta"
- Reported via Slack by Lorenzo Muresan, 2026-08-13

### LINKED ISSUE #188: New Apple Silicon Macs record zero time until someone installs Rosetta by hand
## Business impact

A new Apple Silicon Mac tracks **nothing at all** until Rosetta 2 is installed by hand. The bundled ActivityWatch trackers are x86_64-only, so on a fresh M-series machine every tracker start fails and the agent captures zero events — while the app looks like it is running normally. The user works for hours and then finds the time recorded as Idle/AFK.

This is the **second confirmed occurrence**. `_rosetta_missing()`'s own docstring records the first: Laszlo Fabian Raul's device recorded zero seconds across 2026-07-22 and 07-23. Carmen Lapusan lost roughly 90 minutes on 2026-08-13 setting up a new laptop, then had to raise a support request and back-fill the time manually.

The cost is not only the lost hours. It lands on the worst day — someone's first day on a new machine — and it is indistinguishable to them from the product being broken or from being wrongly marked idle, which is the most sensitive thing a time tracker can get wrong.

## Steps to reproduce

1. Take an Apple Silicon Mac with no Rosetta 2 installed (a clean, freshly-set-up machine).
2. Install and launch BetterFlow, complete sign-in.
3. Work normally for an hour.
4. Look at the timeline in the web app.

## Expected result

Either the trackers run natively, or the app makes it unmissable that it is recording nothing and cannot until Rosetta is installed.

## Actual result

Zero events captured. The timeline shows the whole period as Idle/AFK. The agent detects the cause correctly and logs it clearly, but the running app otherwise looks healthy.

From Carmen's logs (2026-08-13, agent v1.5.122, arm64 build):

```
13:26:01 aw_manager - ERROR - Failed to start bf-data-service: [Errno 86]
         Bad CPU type in executable: .../trackers/darwin/bf-data-service/bf-data-service
13:26:25 aw_manager - ERROR - This Mac is Apple Silicon and Rosetta 2 is not installed,
         so the x86_64 ActivityWatch trackers cannot run. Nothing will be tracked until
         it is installed: run `softwareupdate --install-rosetta`
```

That error repeated 5 times across two app launches and a reboot. `localhost:5600` refused connections throughout (`[Errno 61] Connection refused`), 15 × "Force-restarting tracker stack (server unreachable)". Capture only began at **14:56:12** (`Started bf-idle-tracker`), roughly 90 minutes after first launch, once Rosetta had been installed.

Everything after that is healthy and unrelated: window attribution briefly lost at 16:16 (System Settings frontmost, an excluded app — billing unaffected), screen locked 16:54, machine asleep ~17:03–17:10, and a clean 7–12 events/minute by 17:24.

## Technical details

`src/aw_manager.py:52` pins x86_64 for every platform:

```python
"darwin": f"activitywatch-{AW_VERSION}-macos-x86_64.zip",
```

with `AW_VERSION = "v0.13.2"`. Upstream published no arm64 macOS build for that version, which is why it was pinned — but that is no longer true. Verified against the GitHub releases API today:

| Release | macOS assets |
|---|---|
| `v0.13.2` (pinned) | x86_64 only |
| `v0.14.0b1` | **arm64 only** |
| `v0.14.0b2` | **arm64 and x86_64** |

So a native fix is now available, and `v0.14.0b2` is the one to look at — **not** `v0.14.0b1`, which would silently break every Intel Mac in the fleet for the mirror-image reason. Both are betas, which is the real cost to weigh.

**Detection already works and is not the gap.** `_rosetta_missing()` probes correctly by running a known-good system binary under the x86_64 personality, checks *before* spawning rather than after repeated EBADARCH failures, sets `tracker_download_failed` and `_managed_components_unavailable` so the device reports capture-dead on the heartbeat, and fires `_notify_rosetta_required_once()`. All of that behaved exactly as designed here.

The gap is that **a single toast is the only thing standing between a new user and hours of silent data loss**, and it fires during the noisiest possible moment — first launch on a new laptop, amid permission prompts, sign-in and a browser redirect. Carmen either never saw it or dismissed it with everything else, which is the predictable outcome rather than a mistake on her part.

Options, roughly in order of how much they actually solve:

1. **Ship native arm64 trackers** (`v0.14.0b2`) so the situation cannot arise. Removes the class; costs moving to a beta.
2. **Make the dead state persistent and visible in the app**, not a one-shot notification — a tray state plus a Diagnostics row saying "Not recording — Rosetta 2 required", present for as long as it is true. A one-time toast is the wrong shape for a condition that persists until someone acts.
3. **Block or warn at first-run setup**, before the user starts working, rather than after.
4. **Alert on the capture-dead flags we already send.** The signal reached the server both times and nobody was paged; Laszlo's device sat dead for two days. Internal-tool2 #2298 is currently refining that badge, so the surface exists.

Worth noting the agent binary itself is now correctly arm64 (v1.5.122 arm64 DMG, auto-installed here) — it is only the bundled trackers that are x86_64. Related: #184, same architecture theme from the other direction.

## Environment

- Agent v1.5.122, arm64 build, macOS on Apple Silicon
- Bundled trackers: ActivityWatch v0.13.2, x86_64
- Affected: 2026-08-13 13:26–14:56 EEST (Carmen Lapusan); previously 2026-07-22/23 (Laszlo Fabian Raul)
- Source: user-submitted log bundle `betterflow-logs-20260813-174204`
- Config at the time: `in_process_afk: true`, `stop_external_afk_tracker: false`, `afk_timeout_minutes: 10`

BetterQAFTW:
Two halves of one Mac-architecture surface. Both are agent-side; neither needs a server change to take effect for the user, and the one field that does need a reader is called out at the bottom.

---

### Mechanism

**VERIFIED (by reading; the fault state itself is not reproducible here — see Limits).**

On a clean Apple Silicon Mac with no Rosetta 2, `_rosetta_missing()` returns True, so `_start_locked` (`aw_manager.py:893`) latches `tracker_download_failed` / `_managed_components_unavailable`, fires the one-shot toast, and returns False. Capture never starts, ActivityWatch never comes up, and `_note_aw_unreachable` accumulates until `_escalate_aw_unreachable` fires. That path sets:

```python
self.tray.set_state(TrayState.ERROR, "ActivityWatch not responding")
```

So the tray **was already red for the whole 90 minutes** — it was pointed at the wrong cause. It named a component the user has never heard of, on a machine recording zero seconds, when the fix fits on one line. `_handle_aw_bucket_failure` carried the identical literal, so both watchdogs had the same defect.

The toast that *did* say the right thing is structurally unable to cover this: it is one-shot per process, fires during a new laptop's noisiest minute (permission prompts, sign-in, browser redirect), and `clear_notifications()` wipes it on quit. That was already established in this issue's investigation comment — hence a persistent surface, not a louder toast.

**This is not the same mechanism as the Input Monitoring / TCC blindness** the repo documents elsewhere. That one is a *permission* fault where the process runs and the watcher sees nothing; this is `posix_spawn` returning `EBADARCH` before any tracker process exists. Checked before assuming.

### Fix

- **`AWManager.capture_blocked_remedy()`** — returns the actionable sentence when the Rosetta memo is established `True`, else `None`. Three states, not two: an un-probed `None` earns no remedy, so an unrelated outage on any platform never gets a confident Rosetta instruction. Reads the **memo**, never the probe (`_rosetta_missing()` forks `/usr/bin/arch`, and this is called on the sync cycle while building a tray message).
- **`SyncCoordinator._tracker_fault_message()`** — feeds **both** escalation paths. Routed through one helper rather than patched at each call site: they are the same outage to the person reading the tray, so a remedy on only one leaves a device showing the old wrong sentence depending on which watchdog fired first.
- **`ROSETTA_INSTALL_COMMAND`** spelled once, quoted by the log line, the toast, and now the tray. A typo in one copy is a user typing a command that does nothing on the day they are already recording nothing.

Falls back to the existing text on an absent remedy, a raising manager, or a non-`str`. That last guard is not padding: the message renders verbatim into the tray, so a plain `remedy or fallback` would put `<MagicMock ...>` in front of a user against any test double — mutation M5 below confirms it is load-bearing.

---

## #184 — "who is on the Intel BUILD?" was unanswerable from the wire

Per this issue's last comment, the remaining scope was exactly one thing: the agent sending its **process** architecture as a second field. That is what this does.

### Mechanism

**VERIFIED on real Apple Silicon** (Darwin 25.5.0), running this repo's own helpers under both personalities:

```
### NATIVE:
process_arch=arm64   is_rosetta_translated()=False  true_machine_arch()=arm64

### TRANSLATED (arch -x86_64):
process_arch=x86_64  is_rosetta_translated()=True   true_machine_arch()=arm64
```

`true_machine_arch()` is **identical (`arm64`) in both rows** — by design, because it answers "what silicon is this?". That is precisely why it cannot answer this issue's title: a native Apple Silicon install and an Intel build under Rosetta are byte-identical on the heartbeat today. `platform.machine()` is the only value that moves.

### Negative control for the architecture probe

The brief flagged the known trap — `file <bin> | grep -oE 'arm64|x86_64' | head -1` is wrong because universal binaries list both and first-match wins. That probe is not used anywhere here. The probe used is `platform.machine()`, and it was controlled by running **the same code** against a known-arm64 and a known-x86_64 case:

| case | probe answer |
|---|---|
| native (known arm64) | `arm64` |
| `arch -x86_64` (known Intel personality) | `x86_64` |

Two different answers for two known-different inputs, so the probe measures something real rather than returning a constant. `sysctl.proc_translated` was controlled the same way (`0` native / `1` translated). Both control runs asserted `module.__file__` resolved inside this worktree, so the documented editable-install trap (imports silently resolving to the primary checkout) is ruled out rather than assumed.

One control leg initially failed with `arch: /usr/bin/sysctl isn't executable` — that is `/usr/bin/sysctl` being arm64-only, an artifact of the invocation, not a finding. Re-run through a translated Python it returns `1`.

### Fix

- **`machine_arch.process_arch()`** — `platform.machine()`, forking nothing, with no `""`/undetermined state (a running process always has an architecture). Accepts `translated=` for signature symmetry with its siblings and deliberately ignores it; honouring it would reintroduce the exact Rosetta resolution that makes the field unable to answer its question.
- **`process_arch`** added to `HEARTBEAT_HEALTH_KEYS` **and** to `disclosure_baseline`, each with its written reason (the two tuples are parity-enforced by `test_disclosure_baseline.py`), plus the `CLAUDE.md` device-identifiers section the repo documents as the place to update whenever those keys change.
- Producer in **its own** `try/except` — folded into the existing one, a failure here would silently delete the `machine_arch` that already ships. Mutation M7 confirms that is witnessed.

The pair is the answer, never either half:

| `machine_arch` | `process_arch` | reading |
|---|---|---|
| `arm64` | `arm64` | native Apple Silicon build — fine |
| `arm64` | `x86_64` | **Intel build under Rosetta — the population #184 asks about** |
| `x86_64` | `x86_64` | a genuine Intel Mac — fine |

---

## Proof of failure

Both suites were watched failing against the pre-change implementation, loaded with `git checkout HEAD~1 -- <impl paths>` (never `git stash` — `refs/stash` is shared across worktrees).

`test_rosetta_capture_blocked_message.py` — **6 failed, 4 passed** pre-change:

```
FAILED test_the_manager_reports_the_remedy_when_rosetta_is_the_established_cause
FAILED test_the_manager_withholds_a_remedy_when_rosetta_is_fine
FAILED test_the_manager_withholds_a_remedy_when_the_probe_never_ran
FAILED test_the_remedy_never_forks_a_subprocess
FAILED test_the_tray_names_rosetta_when_that_is_why_capture_is_dead
FAILED test_the_bucket_failure_path_says_the_same_thing
PASSED test_an_ordinary_outage_still_reads_as_before
PASSED test_a_broken_manager_never_costs_the_escalation
PASSED test_a_nonsense_remedy_never_reaches_the_user
PASSED test_an_empty_remedy_falls_back
```

Only the defect-specific cases fail; the four correct-behaviour cases pass in **both** versions. The suite isolates the bug rather than describing the new code.

`test_process_arch_heartbeat.py` fails at import pre-change (`cannot import name 'process_arch'`). Post-change both files are green (22 tests).

### Mutation matrix — one mutant per mechanism, each killed by a **distinct** named test

Anchors asserted unique before mutating, and `__pycache__` cleared between every run (`.pyc` invalidation keys on `(mtime, size)`, so a same-length mutant otherwise runs the *previous* mutant's bytecode and the matrix converges on fiction).

| mutant | killed by |
|---|---|
| M1 remedy drops the command | `test_the_manager_reports_the_remedy_when_rosetta_is_the_established_cause` |
| M2 guard admits the un-probed/fine memo | `..._withholds_a_remedy_when_rosetta_is_fine`, `..._when_the_probe_never_ran` |
| M3 escalate call site back to the literal | `test_the_tray_names_rosetta_when_that_is_why_capture_is_dead` |
| M4 bucket call site back to the literal | `test_the_bucket_failure_path_says_the_same_thing` |
| M5 drop the `isinstance`/empty guard | `test_a_nonsense_remedy_never_reaches_the_user`, `test_an_empty_remedy_falls_back` |
| M6 `process_arch` delegates to `true_machine_arch` | `test_process_arch_reports_the_process_not_the_hardware`, `..._never_consults_rosetta` |
| M7 fold `process_arch` into `machine_arch`'s try | `..._does_not_take_machine_arch_with_it`, `..._never_costs_the_heartbeat` |

**7 mutants, 0 survivors**, control green before and after. M3 and M4 dying to *different* tests is the point — it proves the two call sites are separately witnessed, so a future fix to one cannot silently leave the other behind.

M6 is the agreement-region check: it substitutes the *plausible wrong implementation* (delegating to the arch helper already in the file) rather than deleting code, and the fixtures are chosen at inputs where correct and wrong **diverge**. A suite whose every arch fixture sat at a native value could not have seen it.

Honest caveat on the matrix: zero survivors measures the mutants I thought to write, and I wrote them from my own diff. The enumeration was taken per emitted mechanism rather than per changed line, but it shares an author with the code.

## CI — run locally, because Actions is cost-capped org-wide

`ls .github/workflows/` → `build.yml`, `sync-downloads.yml`, `verify-tracker-pins.yml`.

- **`build.yml` → `test`** (the PR gate): `pytest tests/ -v --tb=short` → **1845 passed, 1 skipped, exit 0**. Baseline 1823 + the 22 added here, so the count moved and my tests demonstrably ran.
- **`build.yml` → `build`**: gated `if: startsWith(github.ref, 'refs/tags/v') || workflow_dispatch` — does not run on a PR. Skipped deliberately, not overlooked.
- **`verify-tracker-pins.yml`**: its `push` path filter includes `src/aw_manager.py`, which this branch modifies, so it **will** fire. Run: all 3 upstream archives fetched and hashed, `All 3 pins verified for AW_VERSION v0.13.2`, exit 0.
- **`sync-downloads.yml`**: `release` trigger only — deploy-side, not applicable.
- **ruff** is not in CI, but every touched source file reports the **identical** finding count as on `origin/main` (0/4/8/2/0), and the two new test files are ruff-clean.

## Limits — what this hardware cannot confirm

Stated plainly rather than left to read as coverage:

1. **The zero-capture state itself is not reproducible here.** This Mac has Rosetta installed (`arch -x86_64 /usr/bin/true` → rc 0) and Rosetta cannot be uninstalled. Every test injects the Rosetta memo. What is verified is that the message the surface carries is *derived from* the blocked state and names the remedy; that a real Apple Silicon Mac without Rosetta *reaches* that branch is inherited from `_rosetta_missing()`, which is unchanged and keeps its own tests.
2. **Needs a clean Apple Silicon Mac to confirm end to end:** that the tray tooltip renders the longer string without truncating the command, and that the new text appears at the moment of the first escalation rather than only after a restart.
3. **CI pins Python 3.11; this ran on 3.14.6.** No 3.11 available on this machine. Nothing here is version-sensitive (`platform.machine()`, string formatting), but the gate itself has not been executed on the pinned interpreter.
4. **`process_arch` has no server reader yet — deliberately.** Verified against internal-tool2 `origin/main`: `machine_arch` resolves in 4 files (heartbeat controller, fleet view, stats, model); `process_arch` in none. Also verified that the heartbeat controller gates on `$request->hasAny([...])` presence checks rather than strict validation, so an unknown key is **ignored, not rejected** — this cannot 422 the fleet. This mirrors exactly how `machine_arch` shipped (agent side in #185/#191/#198, server side later in internal-tool2 #2358). The internal-tool2 half is a follow-up and is the thing that makes the field answer the question on a dashboard.
5. **Not attempted here:** the native arm64 trackers (option 1 in #188, still blocked on upstream shipping only `v0.14.0b*` prereleases) and the heartbeat *cause discriminator* that would let the server badge say "Rosetta 2 missing" instead of "failed download". The latter is a second field with no reader, and this issue's own history is the argument for not shipping two of those at once — it needs the internal-tool2 change beside it.

Multi-session check re-run immediately before opening this: 0 open PRs, no remote branch touching this area, `origin/main` still at `4a1f2ab` (this branch's base), and every deleted line in the diff accounted for as one I replaced.

Closes #188.

Closes #184.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -161,6 +161,17 @@ not the person using it:
   `true_machine_arch()` returned `""`, i.e. the probe never resolved; do not
   coerce that to a string, or "we could not tell" becomes indistinguishable from
   an architecture.
+- `process_arch` (`src/machine_arch.py`) — the CPU architecture of the running
+  **process**, i.e. which BUILD of the agent is installed. Same category as
+  `machine_arch` above: a model of processor, never a person. **Why it is sent
+  separately:** `machine_arch` resolves *through* Rosetta on purpose, so a
+  native arm64 install and an Intel build under Rosetta report the identical
+  value and are byte-identical on the wire. `machine_arch` therefore answers
+  "who is on Intel *silicon*?" while #184 asks "who is on the Intel *build*?" —
+  and only the PAIR separates them (`machine_arch=arm64` with
+  `process_arch=x86_64` is a device sitting on the wrong build). Unlike
+  `machine_arch` it is never `null`: a running process always has an
+  architecture, so there is no "undetermined" state to model.
 
 The serial is shown back to the user in the tray under **Diagnostics > Device
 serial**, next to the Privacy Policy link, and clicking it copies the value.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -169,9 +169,14 @@ not the person using it:
   value and are byte-identical on the wire. `machine_arch` therefore answers
   "who is on Intel *silicon*?" while #184 asks "who is on the Intel *build*?" —
   and only the PAIR separates them (`machine_arch=arm64` with
-  `process_arch=x86_64` is a device sitting on the wrong build). Unlike
-  `machine_arch` it is never `null`: a running process always has an
-  architecture, so there is no "undetermined" state to model.
+  `process_arch=x86_64` is a device sitting on the wrong build). `null` means
+  the same thing it does for `machine_arch` — the architecture could not be
+  determined — and must be handled the same way. A running process essentially
+  always has an architecture, so `null` here should be far rarer than for its
+  sibling, but `platform.machine()` is documented to return `""` when it cannot
+  answer; that is mapped to `null` at the producer rather than shipped as a
+  blank string, because `HEARTBEAT_HEALTH_KEYS` filters by key membership and
+  would otherwise put `""` on the wire.
 
 The serial is shown back to the user in the tray under **Diagnostics > Device
 serial**, next to the Privacy Policy link, and clicking it copies the value.
--- a/src/aw_manager.py
+++ b/src/aw_manager.py
@@ -135,6 +135,13 @@ WINDOW_TITLE_EVENT_LIMIT = 500
 # error reports. start() is retried from the health-check tick, so without this
 # a fail-closed download would toast the user every cycle.
 DOWNLOAD_FAILURE_REPORT_INTERVAL = 3600  # 1 hour
+# The one thing that fixes a zero-recording Apple Silicon Mac, spelled once.
+# Three surfaces quote it — the log line, the toast, and (since #188) the tray
+# state that persists for as long as the fault does. It is the whole remedy, so
+# a typo in any single copy is a user typing a command that does nothing on the
+# day they are already recording nothing; one constant is what stops the three
+# from drifting. The agent cannot run it itself: it needs an admin password.
+ROSETTA_INSTALL_COMMAND = "softwareupdate --install-rosetta"
 # (ops summary, ops fingerprint, user toast) per cause of "this device captures
 # NOTHING". Keyed rather than branched so a new cause is a table row. Both are
 # the same outage to the person using the machine, so both must reach them —
@@ -814,10 +821,55 @@ class AWManager:
             logger.error(
                 "This Mac is Apple Silicon and Rosetta 2 is not installed, so the "
                 "x86_64 ActivityWatch trackers cannot run. Nothing will be tracked "
-                "until it is installed: run `softwareupdate --install-rosetta` "
-                "(needs an admin password, so the agent cannot do it for you)."
+                "until it is installed: run `%s` "
+                "(needs an admin password, so the agent cannot do it for you).",
+                ROSETTA_INSTALL_COMMAND,
             )
         return missing
+
+    def capture_blocked_remedy(self) -> Optional[str]:
+        """One sentence for a surface that PERSISTS, or None.
+
+        Returns the user-facing remedy when this device is recording nothing
+        for a cause the person at the keyboard can actually fix, and ``None``
+        whenever we have not established such a cause. Today that is exactly
+        one cause: Apple Silicon without Rosetta 2.
+
+        **Why this exists.** The Rosetta fault already reached the user twice —
+        a one-shot toast, and a red tray. The toast fires once per process,
+        during the noisiest minute of a new laptop's first launch, and
+        ``clear_notifications()`` wipes it on quit; the tray persisted for the
+        whole outage and said *"ActivityWatch not responding"*, naming a
+        component the user has never heard of. So the fault was reported on a
+        surface nobody could act on and mis-labelled on the surface that lasted.
+        Carmen Lapusan lost ~90 minutes to that on 2026-08-13, after Laszlo
+        Fabian Raul's device recorded zero seconds across 2026-07-22/23. This
+        method is what lets the persistent surface carry the actionable text.
+
+        **Reads the MEMO, never the probe.** ``_rosetta_missing()`` forks
+        ``/usr/bin/arch``; this is called while building a tray message, on the
+        sync cycle, so it consults the cached answer only. The same rule the
+        tray's own ``_arch_menu_item`` and ``serial_menu_row`` carry.
+
+        **Three states, not two.** The memo is ``None`` until something probed:
+
+            True  -> established: no Rosetta, nothing is being recorded
+            False -> established: Rosetta is present, this is some OTHER fault
+            None  -> nobody has asked yet
+
+        Only ``True`` earns a remedy. Treating ``None`` as ``True`` would put a
+        confident Rosetta instruction on every unrelated outage on every
+        platform, including machines that already have it — a guess, on the one
+        surface that exists to stop the user guessing. Fail toward saying
+        nothing, exactly as ``_rosetta_missing()`` itself fails toward
+        attempting the start.
+        """
+        if self._rosetta_missing_cached is not True:
+            return None
+        return (
+            "Not recording — Rosetta 2 required. Open Terminal and run: "
+            f"{ROSETTA_INSTALL_COMMAND}"
+        )
 
     def _notify_rosetta_required_once(self) -> None:
         """Tell the user once per process. The tray state is set by the caller's
@@ -834,7 +886,7 @@ class AWManager:
             outcome = send_notification(
                 "BetterFlow can't track on this Mac",
                 "Rosetta 2 is required. Open Terminal and run: "
-                "softwareupdate --install-rosetta",
+                f"{ROSETTA_INSTALL_COMMAND}",
             )
         except Exception:
             logger.debug("Rosetta notification failed", exc_info=True)
--- a/src/aw_manager.py
+++ b/src/aw_manager.py
@@ -851,21 +851,39 @@ class AWManager:
         sync cycle, so it consults the cached answer only. The same rule the
         tray's own ``_arch_menu_item`` and ``serial_menu_row`` carry.
 
+        **Two conditions, not one, and the second is the one that was missing.**
+        A Rosetta memo of ``True`` says these binaries cannot execute here. It
+        does NOT say this device is recording nothing — a user who installed a
+        native arm64 ActivityWatch themselves has a live server on the port and
+        a memo that will read ``True`` forever. Gating on the memo alone put
+        *"Not recording — Rosetta 2 required"* in front of someone who IS
+        recording, about a component irrelevant to whatever actually failed,
+        the moment any unrelated bucket failure escalated.
+
+        So the remedy also requires ``tracker_download_failed`` — the manager's
+        own latch for "this device captures nothing", set on the same branch of
+        ``_start_locked`` and deliberately NOT set when an external server is
+        attached. ``_managed_components_unavailable`` is the wrong flag for this
+        question: it is true on the recording-via-external device too, by
+        design, because our watchers really are unavailable there.
+
         **Deliberately NOT under ``_lifecycle_lock``**, unlike its sibling
-        ``health_snapshot()`` which reads the capture-dead flags under it. Two
-        reasons, and the second is the one that matters. The memo is write-once
-        and monotonic — whether this Mac can execute x86_64 cannot change
-        without a reboot, so there is no torn read to protect against and a
-        stale read is impossible rather than merely unlikely; the flags
-        ``health_snapshot`` guards are genuinely re-written on every start.
-        And the caller is the tray-message path, which runs while
-        ``force_restart()`` may be holding that lock across process spawns —
-        so taking it here would block a UI update behind tracker teardown to
-        re-read a value that cannot have changed.
+        ``health_snapshot()`` which reads the same latch under it. The caller is
+        the tray-message path, and ``force_restart()`` holds that lock across
+        process spawns — taking it here would block a UI update behind tracker
+        teardown. Both reads are single attribute loads of a bool/tri-state, so
+        there is no torn read; the worst case is a value that was true an
+        instant ago, which is all a tray message can ever claim anyway.
+
+        (Until this commit the argument was "the memo is write-once and
+        monotonic". That stopped being true when ``force_restart()`` began
+        clearing it so an installed Rosetta can be noticed without an agent
+        restart. The conclusion survives; the reason for it did not, and a stale
+        reason is worse than none.)
 
         **Three states, not two.** The memo is ``None`` until something probed:
 
-            True  -> established: no Rosetta, nothing is being recorded
+            True  -> established: no Rosetta, these binaries cannot run
             False -> established: Rosetta is present, this is some OTHER fault
             None  -> nobody has asked yet
 
@@ -877,6 +895,8 @@ class AWManager:
         attempting the start.
         """
         if self._rosetta_missing_cached is not True:
+            return None
+        if not self.tracker_download_failed:
             return None
         return (
             "Not recording — Rosetta 2 required. Open Terminal and run: "
@@ -950,17 +970,44 @@ class AWManager:
             logger.debug("Tracker start refused: capture is suppressed")
             return False
 
+        # Probed BEFORE the Rosetta branch below, not after it. Our managed
+        # binaries being unrunnable says nothing about whether SOMETHING is
+        # capturing on this port: a user who hit the "nothing is being tracked"
+        # wall and installed a native arm64 ActivityWatch themselves is
+        # recording perfectly well. With the probe below the branch, that device
+        # latched the capture-dead flags and never reached the external-attach
+        # path at all, so it reported itself as recording nothing while
+        # recording — and, once capture_blocked_remedy() started reading those
+        # flags, would have been told to install Rosetta over an unrelated
+        # bucket failure. The download-failure path a few lines down has always
+        # asked this question in this order; the Rosetta path did not.
+        server_already_running = self._port_in_use()
+
         # Apple Silicon without Rosetta 2 can never run these binaries. Checked
         # BEFORE spawning rather than after 21 identical EBADARCH failures, and
         # reported rather than retried, because no amount of retrying installs
         # Rosetta. See _rosetta_missing().
         if self._rosetta_missing():
-            self.tracker_download_failed = True
+            # Managed watchers are unavailable either way — these binaries do
+            # not execute on this machine, so nothing here self-heals.
             self._managed_components_unavailable = True
+            if server_already_running:
+                # ...but an external server IS capturing on this port, so this
+                # is not a "capturing nothing" device. Verbatim the rule the
+                # download-failure branch below already applies: attach, log,
+                # and do NOT latch the capture-dead flag or alarm the user.
+                # Telling this person to install Rosetta would name a component
+                # irrelevant to whatever actually went wrong.
+                logger.warning(
+                    "Managed trackers cannot run (no Rosetta 2), but an external "
+                    "server is running on port %s — attaching to it",
+                    self.aw_port,
+                )
+                self._using_external = True
+                return True
+            self.tracker_download_failed = True
             self._notify_rosetta_required_once()
             return False
-
-        server_already_running = self._port_in_use()
 
         binaries_dir = self._get_binaries_dir()
 
@@ -1169,6 +1216,29 @@ class AWManager:
             # Drop external attachment so _start_locked starts our own server
             # if the port is now free after the reap.
             self._using_external = False
+            # Re-probe Rosetta on the way back up. The memo is otherwise written
+            # once per process and cleared nowhere, so a user who read the tray,
+            # ran `softwareupdate --install-rosetta` and watched it finish still
+            # saw "Not recording — Rosetta 2 required" until they thought to
+            # quit and relaunch the agent — and nothing on any surface asked
+            # them to. That is the dead end tray.py's capture_permissions_row()
+            # already names in its own docstring: for a surface whose whole job
+            # is "did my fix work?", still saying "blocked" afterwards is the
+            # failure, not a cosmetic lag.
+            #
+            # Here rather than on a timer because this is the one path that
+            # already exists to rebuild a stack believed dead, and it is exactly
+            # where a device in this state arrives: capture is gone, so the
+            # unreachable watchdog escalates and calls force_restart on every
+            # cycle. Clearing the memo makes the very next start attempt real,
+            # the trackers spawn, and the tray goes green on its own.
+            #
+            # Cost is one `/usr/bin/arch -x86_64 /usr/bin/true` per force
+            # restart, which is a rare escalation path, not the 60s start path
+            # the memo exists to keep clear. A probe that fails still fails
+            # toward "available" (see _rosetta_missing), so the worst case is
+            # one spawn attempt the EBADARCH handler already covers.
+            self._rosetta_missing_cached = None
             return self._start_locked()
 
     def restart_idle_tracker(self, reason: str = "") -> None:
--- a/src/aw_manager.py
+++ b/src/aw_manager.py
@@ -1069,6 +1069,21 @@ class AWManager:
                     self.aw_port,
                 )
                 self._using_external = True
+                # CLEAR the latch, do not merely decline to set it. This branch
+                # returns before the "binaries resolved" clear forty lines down,
+                # so on a Rosetta-missing Mac nothing else ever unsets it — and
+                # a single /info timeout on a healthy external server would
+                # otherwise latch "captures nothing" for the life of the
+                # process, putting a permanent "install Rosetta" in front of
+                # someone who IS recording and reporting a false capture-dead
+                # device to the fleet. Found by walking three cycles with one
+                # transient blip in them, not by reading.
+                #
+                # Safe because of what this branch has just established: a
+                # server ANSWERED /api/0/info on our port. That is the same
+                # evidence the flag's own name asks for, so clearing it here is
+                # the flag being accurate rather than an optimistic reset.
+                self.tracker_download_failed = False
                 return True
             if server_already_running:
                 # Held, but dead. Worth its own line: this device looks
--- a/src/aw_manager.py
+++ b/src/aw_manager.py
@@ -900,6 +900,16 @@ class AWManager:
         question: it is true on the recording-via-external device too, by
         design, because our watchers really are unavailable there.
 
+        **"Attached" means answering ``/api/0/info``, not holding the socket.**
+        The first cut of the condition above rode on ``_port_in_use()``, a bare
+        TCP connect, so ANY process on port 5600 withheld the remedy — and a
+        Rosetta-blocked Mac holding a dead ``bf-data-service`` is precisely a
+        device with a listener and no capture, un-reapable because our binaries
+        have never run there. That returned ``None``, the tray fell back to
+        *"ActivityWatch not responding"*, and the fleet saw
+        ``tracker_download_failed=False``: the exact outage this method exists
+        to name, restored on the exact device it was written for.
+
         **Deliberately NOT under ``_lifecycle_lock``**, unlike its sibling
         ``health_snapshot()`` which reads the same latch under it. The caller is
         the tray-message path, and ``force_restart()`` holds that lock across
@@ -1024,7 +1034,29 @@ class AWManager:
             # Managed watchers are unavailable either way — these binaries do
             # not execute on this machine, so nothing here self-heals.
             self._managed_components_unavailable = True
-            if server_already_running:
+            # `server_already_running` is a TCP connect, and a held socket is
+            # NOT a running tracker. Here — and ONLY here — that distinction
+            # decides whether the user is told anything at all, so this branch
+            # pays for one HTTP ask before granting the carve-out.
+            #
+            # Why this branch and not the download-failure one below: the other
+            # paths use the port answer to decide whether to START something,
+            # and a hung-but-listening server there is recoverable — the
+            # unreachable watchdog escalates to force_restart(), which reaps the
+            # orphan and rebuilds (that is what its docstring describes). On a
+            # Rosetta-missing Mac there is no such recovery: _reap_orphan_
+            # processes is path-scoped to binaries_dir, and OUR binaries have
+            # never executed on this machine, so the listener is un-reapable by
+            # construction. The only decision left on this branch is what to
+            # TELL the person, and getting that wrong cost ~90 minutes once.
+            #
+            # Failing this way is also the safe direction: a false "responding"
+            # withholds the remedy (the regression below), while a false "not
+            # responding" merely offers a Rosetta instruction to a device that
+            # is, in fact, running an ARM ActivityWatch which just failed to
+            # answer /info within 2s — visible, wrong for one cycle, and it
+            # self-corrects on the next probe.
+            if server_already_running and self._server_responding():
                 # ...but an external server IS capturing on this port, so this
                 # is not a "capturing nothing" device. Verbatim the rule the
                 # download-failure branch below already applies: attach, log,
@@ -1038,6 +1070,16 @@ class AWManager:
                 )
                 self._using_external = True
                 return True
+            if server_already_running:
+                # Held, but dead. Worth its own line: this device looks
+                # "started" to every port-level check in the file while
+                # capturing nothing, and nothing here can reap the listener.
+                logger.warning(
+                    "Managed trackers cannot run (no Rosetta 2) and the process "
+                    "holding port %s does not answer /api/0/info — treating this "
+                    "device as capturing nothing",
+                    self.aw_port,
+                )
             self.tracker_download_failed = True
             self._notify_rosetta_required_once()
             return False
@@ -1817,9 +1859,31 @@ class AWManager:
             logger.error(f"Failed to start {name}: {e}")
             return False
 
+    def _server_responding(self) -> bool:
+        """True only if a tracker server ANSWERS ``/api/0/info`` on our port.
+
+        The single HTTP ask that ``_wait_for_server`` polls, lifted out so the
+        two callers cannot drift. This is the difference between *something
+        holds the socket* and *a tracker server is up*, and those are not the
+        same question — ``_port_in_use()`` answers the first and is routinely
+        read as the second. ``force_restart()``'s own docstring names the gap:
+        a ``bf-data-service`` that still holds port 5600 and no longer answers
+        HTTP is exactly what it exists to reclaim.
+
+        One request, one 2s timeout, no retry loop: callers that need to WAIT
+        for a boot own the deadline (``_wait_for_server``); callers deciding
+        what is true right now do not.
+        """
+        url = f"http://localhost:{self.aw_port}/api/0/info"
+        try:
+            req = urllib.request.urlopen(url, timeout=2)
+            req.close()
+            return True
+        except (urllib.error.URLError, OSError):
+            return False
+
     def _wait_for_server(self) -> bool:
         """Wait for tracker server to accept connections."""
-        url = f"http://localhost:{self.aw_port}/api/0/info"
         deadline = time.monotonic() + STARTUP_TIMEOUT
 
         while time.monotonic() < deadline:
@@ -1832,19 +1896,21 @@ class AWManager:
                 )
                 return False
 
-            try:
-                req = urllib.request.urlopen(url, timeout=2)
-                req.close()
+            if self._server_responding():
                 logger.info("Tracker server is ready")
                 return True
-            except (urllib.error.URLError, OSError):
-                time.sleep(0.5)
+            time.sleep(0.5)
 
         logger.error(f"Tracker server not ready after {STARTUP_TIMEOUT}s")
         return False
 
     def _port_in_use(self) -> bool:
-        """Check if something is listening on the tracker port."""
+        """Check if something is listening on the tracker port.
+
+        A bare TCP connect. It proves the socket is held; it does NOT prove a
+        tracker is capturing there — see ``_server_responding()`` for that
+        question, and do not use this one to answer it.
+        """
         import socket
 
         with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
--- a/src/aw_manager.py
+++ b/src/aw_manager.py
@@ -851,6 +851,18 @@ class AWManager:
         sync cycle, so it consults the cached answer only. The same rule the
         tray's own ``_arch_menu_item`` and ``serial_menu_row`` carry.
 
+        **Deliberately NOT under ``_lifecycle_lock``**, unlike its sibling
+        ``health_snapshot()`` which reads the capture-dead flags under it. Two
+        reasons, and the second is the one that matters. The memo is write-once
+        and monotonic — whether this Mac can execute x86_64 cannot change
+        without a reboot, so there is no torn read to protect against and a
+        stale read is impossible rather than merely unlikely; the flags
+        ``health_snapshot`` guards are genuinely re-written on every start.
+        And the caller is the tray-message path, which runs while
+        ``force_restart()`` may be holding that lock across process spawns —
+        so taking it here would block a UI update behind tracker teardown to
+        re-read a value that cannot have changed.
+
         **Three states, not two.** The memo is ``None`` until something probed:
 
             True  -> established: no Rosetta, nothing is being recorded
--- a/src/aw_manager.py
+++ b/src/aw_manager.py
@@ -142,6 +142,14 @@ DOWNLOAD_FAILURE_REPORT_INTERVAL = 3600  # 1 hour
 # day they are already recording nothing; one constant is what stops the three
 # from drifting. The agent cannot run it itself: it needs an admin password.
 ROSETTA_INSTALL_COMMAND = "softwareupdate --install-rosetta"
+# How long a Rosetta answer stands before force_restart() is allowed to re-ask.
+# Installing Rosetta needs no reboot, so the answer CAN change under a running
+# agent and a memo held for the life of the process left the user stuck at "still
+# blocked" with nothing telling them to relaunch. Five minutes because the two
+# costs pull opposite ways: the escalation path that clears this runs every 60s
+# for the whole of a total-capture outage, and re-probing on each one puts a
+# subprocess fork per minute back on precisely the machine the memo protects.
+ROSETTA_REPROBE_INTERVAL = 300.0
 # (ops summary, ops fingerprint, user toast) per cause of "this device captures
 # NOTHING". Keyed rather than branched so a new cause is a table row. Both are
 # the same outage to the person using the machine, so both must reach them —
@@ -538,9 +546,18 @@ class AWManager:
         # nothing, so the backend should know this device is un-self-healing.
         self._managed_components_unavailable: bool = False
         # Tri-state cache for the Apple-Silicon-without-Rosetta check. None =
-        # not probed yet. Rosetta cannot appear or vanish without a reboot, so
-        # this is probed once rather than on every 60s start attempt.
+        # not probed yet. Installing Rosetta needs no reboot, so force_restart()
+        # clears this to let a device that was fixed recover on its own — but
+        # only once per ROSETTA_REPROBE_INTERVAL, because a sustained outage
+        # calls force_restart on every 60s cycle and this memo exists to keep
+        # `/usr/bin/arch` off that path.
         self._rosetta_missing_cached: Optional[bool] = None
+        # monotonic timestamp of the last completed probe, for that interval.
+        self._rosetta_probed_at: float = float("-inf")
+        # The last answer we LOGGED. Re-probing must not reprint the same error
+        # every few minutes for the life of an outage — that is the 60-second
+        # log spam the preflight was written to end, arriving by another door.
+        self._rosetta_logged: Optional[bool] = None
         self._rosetta_notified: bool = False
         # Components whose binary is on disk but cannot EXECUTE (EBADARCH /
         # ENOEXEC). Keeps the capture-dead latch honest when only SOME binaries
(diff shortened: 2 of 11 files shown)


BetterQAFTW:
## Pre-merge audit gate — two-lens review (2026-08-24)

**Do not merge.** One Critical: the #188 half of this PR is inert for the user it was written for.

### CRITICAL — the remedy reaches no user-visible surface; `status_text` is write-only

`TrayIcon.model.status_text` has **zero readers** in `src/`. The producer chain is correct end to end (`main.py:1385`/`1413` → `tray.py:1445-1446`) and then stops. Both render paths ignore it:

- **Menu**: `tray.py:823` renders `_get_status_text(s)`, whose ERROR branch (`tray.py:1089-1090`) returns the constant `"Error"`. `s["status_text"]` is in the snapshot and never referenced.
- **Tooltip**: `icon.title` is assigned only from `_update_tooltip` (`tray.py:1558`, `1580`) as `"… - Today: Xh Ym active"`.

What the affected user actually sees: `App status: Error`, `ActivityWatch: Not running`, tooltip showing hours. The word "Rosetta" and the command appear nowhere.

**This also corrects the PR's own diagnosis.** The body says the tray showed *"ActivityWatch not responding"* for 90 minutes — that literal was never rendered either; `git log -S'status_text'` shows the field write-only since initial release. So #188's real defect is that the persistent surface names **no cause at all**, and routing a better string into the same dead field cannot change what the user sees.

**Why the suite is blind to it (Phantom 7, one level up):** `test_rosetta_capture_blocked_message.py:82` sets `c.tray = MagicMock()` and `:89` asserts on `coord.tray.set_state.call_args[0][1]`. The file's section header claims to be witnessing *"the tray message the user actually sees"* — but the asserted consumer is a mock argument one layer above the real consumer (`TrayIcon._get_status_text`), which does not read it. M3 and M4 both die against that mock and prove nothing about rendering.

**Fix direction**: render it — `_get_status_text`'s ERROR branch returning `s["status_text"] or "Error"`, or a top-level menu row mirroring the `⚠️ Input tracking off — Fix Permissions` item at `tray.py:826-832`, which is the pattern that demonstrably renders. Witness with a test driving the REAL `TrayIcon`: `test_tray_state_transitions.py:116-123` already calls `_get_status_text` without a live pystray backend, so there is no hardware excuse here.

### Important 1 — `capture_blocked_remedy()` never checks that capture is blocked

`aw_manager.py:879` is the whole condition: `if self._rosetta_missing_cached is not True: return None`. Ordering in `_start_locked` latches that memo (`:957`) BEFORE the port check (`:963`), so on a machine where the user installed a native arm64 ActivityWatch themselves, the memo is `True` while capture is healthy. An unrelated bucket failure then renders **"Not recording — Rosetta 2 required"** on a machine that IS recording, about a component irrelevant to the failure. `test_the_bucket_failure_path_says_the_same_thing` currently pins that as correct. Gate the remedy on the capture-dead state the manager already computes, not on the Rosetta memo alone.

### Important 2 — the memo never invalidates, and the remedy omits the step that makes it work

`_rosetta_missing_cached` is written once per process and reset nowhere outside tests, so after the user runs the command nothing changes until the agent restarts — and neither the tray copy nor the toast says to restart (only the log carries the admin-password caveat). The repo's own sibling surface states this as a requirement, `tray.py:353-359`: *"for a surface whose whole job is 'did my fix work?', saying 'still blocked' afterwards is the dead end this exists to close."* Add the restart step to the remedy sentence, and/or re-probe on `force_restart` so the surface can go green. Note both lenses reached this independently.

### Important 3 — `process_arch()` can return `""`, and five places promise it cannot

`machine_arch.py:504` is `return machine or platform.machine()`, and the stdlib documents `platform.machine()` returning `""` when undeterminable (reachable on Windows with a scrubbed service environment). The sibling is mapped and the new field is not, five lines apart: `main.py:2093` `arch or None` vs `main.py:2108` bare. `HEARTBEAT_HEALTH_KEYS` filters by membership, so `""` reaches the wire. The "never null" claim appears in `machine_arch.py`, `main.py:2100`, `bf_client.py`, `disclosure_baseline.py` and `CLAUDE.md` — and the deferred internal-tool2 reader will be written against it. No fixture exercises the fallback branch, so it is unwitnessed by construction. Fix: `process_arch() or None`, or delete the claim from all five places.

### Resolved outside this PR

The **untracked reader** finding is closed: internal-tool2 #2373 now tracks persisting `process_arch` (verified against internal-tool2 `origin/main` that the key is dropped on receipt — not merely unread, so it cannot be backfilled). `Closes #184.` may stand.

### Verified clean — worth recording

- **The fleet-facing risk does not exist.** `AgentHeartbeatController.php` on `origin/main` has no `validate`/FormRequest/Validator, reads every field via `input()`/`has()`, `hasHealthTelemetry()` is a `hasAny()` presence check, and the global middleware stack has no schema layer. An unknown `process_arch` key cannot 422 the fleet. **Read from `origin/main` via the API, not the local clone** — that checkout is on a branch predating `machine_arch` entirely, so a local grep would have answered "no reader" and been the stale-checkout trap.
- **No third escalation site.** Every other `set_state(TrayState.ERROR, …)` is sync-failure or login, a different class; `ROSETTA_INSTALL_COMMAND` and `"ActivityWatch not responding"` are each spelled once in `src/`.
- **The three-state `None` guard is correct** (`is not True`, so an un-probed `None` cannot fall through) and the `isinstance(remedy, str)` guard genuinely earns its place against MagicMock fixtures.
- **The arch-pair fixture sits at the divergence point**, not in an agreement region: `machine="x86_64", translated=True` is where `process_arch` and `true_machine_arch` disagree, and M6 is a real substitution mutant.
- Test quality is above par generally — producer and consumer witnessed separately elsewhere, real subprocess-fork negatives, `None`- and `False`-memo controls.

### Optional, logged not fixed

- Redundant f-string wrappers around a bare constant (`aw_manager.py:883`, `901`).
- The remedy SENTENCE is near-duplicated across tray and toast (the command is unified; the sentence is not), and only the log mentions the admin password — which on an MDM-managed laptop is the whole answer.
- `tray.py:310` inlines what `process_arch()` now names, for the same question.
- No DPO sign-off recorded for a new `disclosure_baseline` entry, though the file's header requires one. Precedent commit `fed3862` added `machine_arch` the same way, so this is house practice rather than a regression — flagging so the ping is a decision rather than an omission.

BetterQAFTW:
## Fix round — all four findings addressed, per-item evidence

Branch `fix/mac-arch-detection-188-184` @ `0b7db80`. Four commits on top of `210ee1f`. Not merged.

### CRITICAL — the remedy now renders

Chose **(a)**, `_get_status_text`'s ERROR branch returning `s["status_text"] or "Error"`, over the new top-level menu row. Reason: (a) is a wording change to a row that already renders on an already-firing surface, which is what the parent commit claims to be; (b) is new UI needing its own placement, enablement and dismissal rules. `7518ad6`.

The read sits *inside* the ERROR branch rather than at the top with its siblings — only the states this branch covers write the field, and a top-level read would make every caller supply a key it has no business knowing (the partial dict in `test_tray_state_transitions.py` would raise).

**A second defect fell out of making the field render**, found by running the new test rather than by reading: the model seeds `status_text` to `"Starting..."`, so a text-less escalation drew `App status: Starting...` under a red icon. `set_state` now clears on ENTERING a fault, mirroring the recovery clear. `ERROR -> ERROR` deliberately keeps the sentence — that is one outage escalating again, and blanking it would darken the surface on the second watchdog tick.

Witnessed against the **real `TrayIcon`** driving **`_create_menu`**, asserting on the label list the menu is actually built from — not on a mock argument:

| mutant | kills |
|---|---|
| M1 ERROR branch back to the constant | `test_the_error_status_carries_the_sentence_it_was_given`, `..._reads_the_snapshot_too`, `test_the_menu_row_the_user_reads_names_the_remedy`, `test_a_repeat_escalation_keeps_the_sentence...` |
| M2 branch kept, **render path removed** | `test_the_menu_row_the_user_reads_names_the_remedy` (`App status: Error`), `test_recovery_stops_rendering_the_stale_sentence` |
| M3 entry-clear disabled | `test_an_ordinary_error_still_reads_as_before`, `test_a_previous_states_message_never_reappears_under_a_fault` |

M2 is the one that matters: the four branch-level tests survive it, the two render-path tests do not. That is exactly the class of defect this finding was.

### Important 1 — the remedy is owed only where capture is dead

`930ef58`. Three changes, one fault:

1. **`_start_locked` probes the port BEFORE the Rosetta branch.** The download-failure branch forty lines down has always asked in this order, with the comment saying why; the Rosetta branch was written later and did not. So the external-server device latched the capture-dead flags and never reached the attach path at all — it reported capturing nothing *while capturing*, which is the defect the remedy then inherited.
2. **`capture_blocked_remedy()` also requires `tracker_download_failed`** — the manager's own "captures nothing" latch, deliberately not set on the external-server device per (1). `_managed_components_unavailable` is the **wrong** flag here: it is true on the recording device too, by design.
3. `_managed_components_unavailable` still set either way — those watchers genuinely are unavailable and the fleet needs to know the device cannot self-heal.

**`test_the_bucket_failure_path_says_the_same_thing` changed, and this is an existing assertion changed on purpose.** Through a `MagicMock` manager it pinned "a bucket failure renders the Rosetta remedy" *unconditionally* — the defect. It now drives a real manager in the state the sentence describes, and a new sibling covers the recording device asserting the generic fallback. Not a check weakened to get green: the old one was wrong.

Two existing preflight tests now pin `_port_in_use`. They reached it for the first time via (1) and were otherwise **answering differently per machine** — this laptop has a live server on 5600, which is how the change was seen working before any test asserted it.

| mutant | kills |
|---|---|
| M4 revert the capture-dead gate | `test_the_manager_withholds_a_remedy_from_a_device_that_is_recording`, `test_a_bucket_failure_on_a_recording_device_names_no_remedy` |
| M5 port probe back below the branch | `test_an_external_server_is_attached_to_rather_than_declared_dead` |

### Important 2 — the memo dead-end

Chose **re-probe over restart wording**: it closes the loop instead of documenting it. `force_restart` clears the memo, so an installed Rosetta is noticed without an agent restart. `930ef58` + `0b7db80`.

**That first cut was wrong and the follow-up commit says so.** `_note_aw_unreachable` returns True on *every* cycle once its 180s grace period is up — only a reachable ActivityWatch resets the clock — so a device capturing nothing force-restarts **every 60 seconds** for the whole outage. An unconditional clear forked `/usr/bin/arch` once a minute on exactly the machine the memo protects, and reprinted the same six-line error each time: the "21 identical failures at 60-second intervals" pattern the preflight replaced, wearing a different hat. Found by reading the escalation cadence after the fact, not by a failing test.

So: `ROSETTA_REPROBE_INTERVAL` (300s) bounds it, and `_rosetta_missing` logs the **answer changing** rather than the probe running. The `False`-after-`True` transition gets its own line — the only record that the user's install landed, which is what support reads these logs to answer.

The `capture_blocked_remedy` lock docstring is **rewritten, not left**: its "write-once and monotonic" argument is now false. The conclusion survives (single attribute load, no torn read, must not block a UI update behind `force_restart` holding the lifecycle lock); a stale reason is worse than none.

| mutant | kills |
|---|---|
| M6 `force_restart` stops clearing | `test_installing_rosetta_is_noticed_without_an_agent_restart` |
| M9 rate limit removed | `test_the_re_probe_is_rate_limited_off_the_sixty_second_escalation` |
| M10 log-once collapsed to log-every-probe | `test_a_persisting_outage_logs_the_cause_once_not_once_per_re_probe` |

The rate-limit test asserts **both ends** of the boundary — inside the window the probe must not re-fork, past it it must, or the first assertion is equally satisfied by a memo that never re-probes at all. Its first draft advanced `5 x 60 = exactly 300`, landed on the `>=` and failed, which reads exactly like a broken rate limit rather than a fixture off by one; it now asserts its own precondition.

### Important 3 — `process_arch()` can return `""`

`5242380`. `telemetry["process_arch"] = process_arch() or None`, matching the sibling five lines up. `HEARTBEAT_HEALTH_KEYS` filters by key membership, not truthiness, so `""` shipped.

The claim is corrected in the docs rather than deleted: null here means what it means for `machine_arch`, and should be far rarer — which makes it likelier to ship unhandled, not less costly when it does.

**The review listed five sites; there are four.** `disclosure_baseline.py`'s `process_arch` entry declares the privacy category and makes no nullability claim, so it needed no change. Changed: `machine_arch.py` docstring, `main.py:2100` comment, `bf_client.py` allowlist comment, `CLAUDE.md`.

The branch was unwitnessed **by construction**: every existing test either passes `machine=` explicitly or patches `process_arch` wholesale — which is what makes them meaningful on an ubuntu runner and exactly what hides the right-hand side of that `or`. Four new tests patch `src.machine_arch.platform.machine` and drive the real function, including a **positive control** (a real arch falls through) so a mutant deleting the fallback cannot leave the rest passing on `None`.

| mutant | kills |
|---|---|
| M7 revert the null mapping | `test_a_blank_architecture_reaches_the_wire_as_null_not_as_a_string`, `test_the_pair_maps_an_undeterminable_arch_the_same_way_on_both_halves` |
| M8 delete the `platform.machine()` fallback | `test_process_arch_falls_through_to_the_real_probe`, `test_an_undeterminable_architecture_is_empty_not_a_lie` |

### CI — run locally, both jobs (GitHub Actions is capped; nothing runs on its own)

```
pytest tests/ -v --tb=short     ->  1869 passed, 1 skipped, exit 0
                                    (branch point 210ee1f: 1845 passed, 1 skipped)
python scripts/verify_tracker_pins.py
                                ->  exit 0, all 3 pins verified for AW_VERSION v0.13.2
```

`verify-tracker-pins` triggers on `src/aw_manager.py`, which this branch modifies, so it is not optional here. It fetched all three real upstream archives.

The +24 reconciles exactly: 12 new tray render tests, 6 preflight, 2 rosetta-message, 4 process-arch. Whole-tree `ruff` count **unchanged at 180** (a `RECORDING` in a test name added an `N802` to a previously-clean file and was renamed).

Caveat, stated rather than buried: local Python is **3.14.6**, CI pins **3.11**. `requires-python` is `>=3.10` and nothing here is version-sensitive, but I could not run the suite under 3.11 on this machine — no 3.11 interpreter and no pytest under the 3.12 that is present.

### Not addressed (the review logged these as optional)

Redundant f-string wrappers (`aw_manager.py`); the near-duplicated remedy *sentence* across tray and toast (the command is one constant, the sentence is not); only the log mentioning the admin password — which on an MDM laptop is the whole answer; `tray.py:310` inlining what `process_arch()` now names; the missing DPO sign-off for the new `disclosure_baseline` entry.

BetterQAFTW:
## Round 2 — fresh-context refutation. DO NOT MERGE.

The fix round closed the Critical and all three Importants, and items 1–4 hold under attack (evidence below). But fix #2 introduced a **new Critical regression** on the exact device class this PR targets, and a test now certifies it as correct — so it will not surface on its own later.

Method note worth recording: machine load reached 54 during this review (another writer agent), so the reviewer ran **no suites** — per `cross-repo-safety.md` Rule 7a a suite under that load produces timeouts that read as code bugs. It instead wrote deterministic single-process probes driving the REAL `TrayIcon._create_menu` and REAL `AWManager._start_locked`/`capture_blocked_remedy`, stubbing only `_port_in_use` and the `/usr/bin/arch` fork. Every number below is quoted from those runs.

### CRITICAL — the reordering withholds the remedy whenever ANYTHING answers port 5600

`_port_in_use()` (`aw_manager.py:1846-1856`) is a bare TCP connect. `:1027-1040` treats any listener as "an external server IS capturing", sets `_using_external`, returns `True`, and skips `tracker_download_failed = True`. The remedy's new second condition (`:932`) then returns `None`, and `_tracker_fault_message()` falls back to `_GENERIC_TRACKER_FAULT` — `"ActivityWatch not responding"`, the literal this PR exists to stop showing.

Real `_start_locked`, Rosetta memo `True`, only the port answer varied:

```
F1  nothing on 5600            -> _start_locked False, tracker_download_failed True,
                                  remedy = "Not recording — Rosetta 2 required. …"
F2  SOMETHING answers TCP 5600 -> _start_locked True,  tracker_download_failed False,
                                  _managed_components_unavailable True, remedy = None
```

**A regression:** pre-fix (`210ee1f:src/aw_manager.py:957-961`) the branch was unconditional.

**Reachable, and the repo documents the state:** `force_restart`'s docstring (`:1230-1233`) names "a `bf-data-service` that still holds port 5600 but no longer answers HTTP" as what it exists to reclaim. On a Rosetta-missing Mac that listener is un-reapable by construction — `_reap_orphan_processes` is path-scoped to `binaries_dir` (`:1152-1176`) and our binaries have never executed.

**The steady state is contradictory and invisible on both surfaces:** `_start_locked()` returns True (started fine) while `is_running()` fails `/info`, so the tray reads "ActivityWatch not responding" forever, AND `tracker_download_failed` is False so the fleet cannot see the device either.

**No test can see it and one certifies it:** `test_aw_manager_rosetta_preflight.py:158` patches `_port_in_use` to `True` and asserts `capture_blocked_remedy() is None` under the docstring "This person is recording." Every fixture SUPPLIES the boolean production COMPUTES — Phantom 14. Change that assertion and say in the commit that it was wrong.

**The right probe already exists in the file:** `_wait_for_server` hits `/api/0/info` (`:1822`). Gate the carve-out on an `/info` response rather than a TCP connect — one `urlopen(timeout=2)` on a path that runs at most once per 60s — and witness it with a fixture where the port answers and `/info` does not, the case no current test can express.

### IMPORTANT — the entry-clear guards only half its inputs

`tray.py:1469-1488` clears only when `state in faulted and previous_state not in faulted`, and `faulted` contains BOTH `ERROR` and `QUEUE_WARNING`. So a text-less `ERROR` entered FROM `QUEUE_WARNING` matches neither clear, and the previous message survives into the red row — the exact harm the new comment at `:1479-1482` describes. Real `TrayIcon`, real `_create_menu`:

```
B  ERROR(remedy) -> QUEUE_WARNING("Queue 92% full") -> ERROR(no text)
     -> 'App status: Queue 92% full'      # queue text under a tracker fault
C  ERROR(remedy) -> QUEUE_WARNING(no text) -> ERROR(no text)
     -> 'App status: Not recording — Rosetta 2 required. …'   # stale remedy resurrected
D  control: ERROR(remedy) -> SYNCING -> ERROR(no text) -> 'App status: Error'   # guarded half works
```

Not live today (`_select_tray_state` never returns `ERROR`, and every `set_state(ERROR, …)` passes a message) — but that is the SAME precondition given for adding the guard at all, so the missing half earns the same line. `diagnosis-discipline.md` Rule 3: a guard you write has two ends, and you guard the one you were reasoning about. Fix: `elif state in faulted and previous_state != state:` — which also preserves the deliberate `ERROR → ERROR` retention.

### IMPORTANT — the same menu still names a second, false cause

```
App status: Not recording — Rosetta 2 required. Open Terminal and run: softwareupdate --install-rosetta
ActivityWatch: Not running     <- consistent
API: Unreachable               <- FALSE
```

`_check_api_status` (`tray.py:1168-1171`) derives purely from tray state (`state not in (QUEUED, QUEUE_WARNING, ERROR)`). `ERROR` is now a TRACKER fault, but the API is demonstrably reachable — the heartbeat is how `tracker_download_failed` reaches the fleet. Pre-existing, but **this PR creates the collision**: while the row said a vague "Error", "API: Unreachable" read as elaboration; now the top row makes a specific actionable claim and the first submenu a worried user opens contradicts it. Stop deriving `API:` from `TrayState.ERROR`, or split the tracker fault off `ERROR`.

### HOLDS — tried and could not break

- **The Critical fix genuinely renders.** The test drives the real `_create_menu` with `Item` patched to record labels and asserts exactly one `App status:` row plus content, so removing the read OR deleting line `823` reddens it. Probe A shows the remedy in the real label list; probe E confirms the `Starting...` seed defect is closed.
- **The re-probe arithmetic is right at both boundaries.** Real clock walk at a 60s cadence, counting actual forks: one fork per 300s exactly, no thrash on a stable answer, guaranteed convergence, worst case 300s + one escalation cycle. `_rosetta_probed_at` is stamped only on a real probe, so the cached path cannot starve the interval — the failure mode looked for specifically. The test asserts both ends AND its own precondition, so the disclosed exactly-300 fixture bug cannot recur silently.
- **The rejected alternative was correctly rejected** — verified rather than trusted: on the external-server device `_managed_components_unavailable` really is `True`, so gating on it would fire on a recording machine.
- **The count correction is right: FOUR sites, not five.** `disclosure_baseline.py:585-592` declares the privacy category and makes no nullability claim.
- **No tearing** between the two attributes: `_stop_locked` does not touch `tracker_download_failed`, and both `_tracker_fault_message()` call sites run on the sync thread strictly after `force_restart()` returns.

### Optional, logged

- `_rosetta_missing()` fails toward `False` on a probe exception, so a device discovering the fault via `EBADARCH` has `tracker_download_failed=True` with the memo `False` and gets no remedy; the docstring's "exactly one cause" does not cover it. Pre-existing gate.
- The re-probe interval is inert off Apple Silicon (early return precedes the timestamp stamp), so the memo re-clears on every `force_restart` there. Harmless — that branch forks nothing — but a reader will assume otherwise.
- The tooltip never carries the cause; the remedy requires a click.
- **UNVERIFIED:** the rendered row is 103 chars with no truncation anywhere in our code, but a real `NSMenu` was not rendered, so macOS clipping is unchecked.

BetterQAFTW:
## Round 3 — the regression is closed, per-item evidence. Not merged.

Branch `fix/mac-arch-detection-188-184` @ `c0d35bc`, three commits on top of `0b7db80`.
Suite **1879 passed, 1 skipped, exit 0** (round-2 baseline 1869; +10 reconciles exactly to the 10 tests added below).

### CRITICAL — the carve-out now requires an answer, not a socket

`cd38510`. `_server_responding()` — the single `/api/0/info` ask `_wait_for_server` already polled, lifted out so the two callers cannot drift — gates the Rosetta external-server branch. `_wait_for_server` calls it instead of inlining the request.

**Scoped to that one branch deliberately.** The other three `server_already_running` reads decide whether to START something, and a hung listener there is recoverable: the unreachable watchdog escalates to `force_restart()`, which reaps and rebuilds. On a Rosetta-missing Mac there is no such recovery — `_reap_orphan_processes` is path-scoped to `binaries_dir` and our binaries have never executed — so the only decision left on that branch is what to tell the person. The asymmetry is written at the branch.

Failure direction is the safe one: a false "responding" withholds the remedy (the regression); a false "not responding" offers a Rosetta instruction for one cycle to a device running an ARM ActivityWatch that missed a 2s budget, and self-corrects on the next probe.

**Two assertions were wrong, both for the same reason — every fixture SUPPLIES a boolean production COMPUTES (Phantom 14).**

1. `test_an_external_server_is_attached_to_rather_than_declared_dead` pinned `_port_in_use=True` alone under the docstring *"This person is recording."* A TCP connect cannot establish that premise. **The FIXTURE was the defect, not the assertion** — deviation from the brief's "change that assertion", stated because flipping it would make it wrong in the other direction: with `_server_responding` pinned True the docstring is true of the device the fixture builds, and the case it was wrongly covering gets its own test.

2. **Found by running that new test, not by reading it:** patching the `_rosetta_missing` METHOD skips the memo write the real probe performs, so `_rosetta_missing_cached` stayed `None` — and `capture_blocked_remedy()` returns `None` on a `None` memo by its own first line. That test's "THE assertion" passed identically against the bug, against the fix, and against no implementation at all. Witnessed directly: with the capture-dead gate deleted (M14), the **old** fixture is green and the **new** one reddens.

| mutant | kills |
|---|---|
| M11 `/info` gate reverted to `if server_already_running:` | `test_a_held_port_that_answers_nothing_is_not_a_recording_device`, `test_the_carve_out_asks_over_http_not_over_tcp` |
| M12 `_wait_for_server` re-inlines the request | `test_server_responding_is_the_http_ask_wait_for_server_polls` |
| M13 `_server_responding` returns True for every input | `test_server_responding_is_the_http_ask_wait_for_server_polls` (the negative control) |
| M14 capture-dead gate deleted | `test_an_external_server_is_attached_to_rather_than_declared_dead` + 2 in `test_rosetta_capture_blocked_message.py` — **and green under the pre-fix fixture, which is the point** |

### SELF-REVIEW — the fix opened a new permanent-wrong, closed before pushing

`c0d35bc`. `/info` can time out on a **healthy** server. That one cycle latches `tracker_download_failed`, and the Rosetta branch returns before the "binaries resolved" clear forty lines down, so on a Rosetta-missing Mac nothing else ever unsets it. Walked three cycles with one blip:

```
cycle 0  started=False  download_failed=True  using_external=False  remedy=YES
cycle 1  started=True   download_failed=True  using_external=True   remedy=YES
cycle 2  started=True   download_failed=True  using_external=True   remedy=YES
```

A device recording perfectly via an external server, showing a permanent *"Not recording — Rosetta 2 required"* and reporting itself capture-dead to the fleet: round 1's Important 1 arriving through a transient instead of a configuration, inside the fix for round 2's Critical. The attach branch now CLEARS the latch rather than declining to set it — safe on the evidence the branch just established, a server ANSWERED `/info`. The blip cycle itself is unchanged on purpose; softening it re-opens the Critical. M20 (clear deleted) kills `test_a_transient_info_failure_does_not_latch_for_the_life_of_the_process`.

### IMPORTANT — the entry clear guards both ends

`10f4fc6`. `previous_state != state`, not `previous_state not in faulted`. Both refutation sequences now have a test, and `test_the_guarded_half_still_works` keeps the original door explicit.

| mutant | kills |
|---|---|
| M15 back to the membership test | `test_a_queue_message_does_not_survive_into_a_tracker_fault`, `test_a_stale_remedy_is_not_resurrected_by_a_hop_through_the_other_fault` |
| M16 clear on EVERY fault entry (kills the retention) | `test_a_repeat_escalation_keeps_the_sentence_it_already_carries`, `test_a_repeat_queue_warning_keeps_its_own_sentence` |
| M17 `previous_state != TrayState.ERROR` special-case | `test_a_repeat_queue_warning_keeps_its_own_sentence` **only** |

M17 is why the QUEUE_WARNING diagonal earns its own test: a special-case satisfies every other test in the file and silently blanks that one. It is also the single test here asserted on the MODEL rather than the menu row — `_get_status_text`'s QUEUE_WARNING branch returns the constant `"Offline (queue full)"` and reads no field, so pinning a row there would be a lie. Said in the docstring rather than hidden.

### IMPORTANT — the menu no longer names a second cause

`10f4fc6`. ERROR dropped from `_check_api_status`. Every producer of that state was read, and none implies an unreachable API — the two tracker escalations reach the backend on the same cycle to report the fault, `_report_sync_failure` asks `bf.is_reachable()` first and routes unreachable to `QUEUED("Offline")`, and the login handlers get a server that answers. The reasoning is in the docstring, not just the commit.

| mutant | kills |
|---|---|
| M18 ERROR back in the tuple | `test_a_tracker_fault_does_not_also_claim_the_api_is_unreachable` |
| M19 `return True` (the over-correction) | `test_the_queue_states_still_report_the_api_as_unreachable` |

### The supplied-boolean class — swept, and it does not end at this PR

`_port_in_use` is supplied at **6 test sites**, and at every one the surrounding prose calls it *"external server healthy / capturing / running"* — a claim a TCP connect cannot support. Two production sites hold the same conflation and are **untouched by this branch**. Probed rather than read, varying only whether `/info` answers:

```
_start_component EBADARCH carve-out (aw_manager.py:1851)
  port held, /info ANSWERS  -> tracker_download_failed = False
  port held, /info DEAD     -> tracker_download_failed = False   <- identical

check_health external branch (aw_manager.py:1325)
  port held, /info ANSWERS  -> check_health = True
  port held, /info DEAD     -> check_health = True               <- identical
```

**Not fixed, deliberately.** Both are pre-existing, neither is reachable through the Rosetta preflight (it returns before `_start_component`), both are recoverable via the `force_restart` reap on a machine whose binaries do execute, and the EBADARCH one would add a blocking HTTP call under `_lifecycle_lock` on a per-watcher hot path. This is round 3 of 3 and widening scope here is how the branch acquires its next defect. Worth its own issue; `_server_responding()` now exists for whoever takes it.

`_rosetta_missing` is supplied at 4 sites; `test_rosetta_capture_blocked_message.py:121` sets the memo directly, which is the honest form. The two in this file are now fixed.

### CI — run locally, both jobs (Actions is capped; nothing runs on its own)

```
pytest tests/ -v --tb=short  ->  1879 passed, 1 skipped, exit 0   (baseline 0b7db80: 1869)
ruff check .                 ->  183 findings at BOTH 0b7db80 and c0d35bc
```

Ruff measured the same way on both trees; the whole diff between the two finding sets is line-number shifts of pre-existing findings, no new ones. (Round 2 quoted 180 — a different count method, not a regression; the base-vs-head comparison above is the load-bearing one.)

`verify-tracker-pins` triggers on `src/aw_manager.py`, so it was run in full. **Updated — it has now completed green** (it was flagged pending in the first version of this comment while the three ~100 MB archives downloaded over a 332 KB/s link):

```
[darwin]  OK e62a76c0ec3c0e69d58ba207bb8da6d8d47d0c7ad1bc871ddf702168f291cf5b
[linux]   OK 8f62b10babf8a8f108cbdf7267c02fbc1ce2a970fa9535f230b3416b803e3360
[windows] OK a067fa765678a411991826c4da811fd2d8ca260c2db9d6d897957565b61c369f

All 3 pins verified for AW_VERSION v0.13.2.
PINS EXIT: 0
```

`PINS EXIT: 0` is the verifier's own status, captured immediately after it rather than at the end of the chain — the task harness separately reported "exit code 0" for the whole compound command, which ends in `tail` and would have said 0 either way. The content and the code agree; only the content is evidence.

Independently, my diff touches no pin: `AW_VERSION`, `RELEASE_SHA256` and `RELEASE_ASSETS` are byte-identical to `0b7db80`, verified by comparing the blocks.

Caveat carried forward unchanged: local Python is 3.14.6, CI pins 3.11.

### Logged, not fixed

- `_check_aw_status` has the sibling of the API-row defect — it renders `ActivityWatch: Not running` for a LOGIN error too. Left alone: the refutation verified it consistent for the tracker-fault case this PR creates, and `STARTING`/`WAITING_AUTH` in that tuple are genuinely pre-tracker states.
- Everything the round-2 review logged as optional remains optional and untouched.
- The `/info` gate adds up to 2s under `_lifecycle_lock`, on a path that already tolerates `_wait_for_server`'s 10s, and only when Rosetta is missing AND the port is held.

Issue #216, opened 2026-08-24
Bumping the ActivityWatch pin will not remove the Rosetta requirement — the preflight asks the wrong question


BetterQAFTW:
## Business impact

The obvious next step after #188 is to bump ActivityWatch to a version that ships native Apple Silicon binaries, so new Macs stop needing Rosetta. That alone will not work, and the failure mode is the same silent one #188 was about: a clean Apple Silicon Mac with no Rosetta would still refuse to start capture, still record zero time, and now do it while shipping binaries that would have run fine.

## The mechanism

_rosetta_missing() gates every tracker start and asks whether this Mac can execute x86_64, by running /usr/bin/arch -x86_64 /usr/bin/true. That is the right question only while the bundled trackers are x86_64. Once they are native arm64 it is the wrong question: on a clean Apple Silicon Mac the probe still answers missing, the guard still returns False, and native binaries are never spawned.

The predicate needs to become "is the BUNDLED BINARY x86_64 AND is Rosetta missing", resolved from the shipped binary's architecture or from a build-stamped constant — not from what the host happens to support.

## Two structural blockers in the way (verified here)

**1. The version pin exists twice and only one copy is guarded.**

    src/aw_manager.py:47      AW_VERSION = "v0.13.2"
    scripts/download_aw.py:15 AW_VERSION = "v0.13.2"

scripts/verify_tracker_pins.py imports AW_VERSION, RELEASE_ASSETS, RELEASE_SHA256 from src.aw_manager (line 25), so it covers that copy only. Nothing asserts the two agree. scripts/download_aw.py is what .github/workflows/build.yml uses to fetch binaries, so a drift means the build downloads one version while the runtime pins another — and the guard stays green. Collapse to one definition, or add an equality assertion; do not leave two.

**2. RELEASE_ASSETS, RELEASE_SHA256 and _get_platform_key are keyed on platform with no architecture dimension.** Shipping both macOS arm64 and macOS x86_64 needs that key to gain an arch, which is a schema change to the pin table and to verify_tracker_pins.py, not a value edit.

## Upstream readiness — reported by betterqa-48, not independently verified here

Measured by another session on an M-series Mac and posted to #188 (see that issue's comment thread for the full detail including SHA-256 for all four archives):

- v0.14.0b4 contains all three binaries we rename in every archive (macOS arm64 and x86_64, Windows, Linux), in the same nested layout extract_binaries() already handles.
[...]

Issue #215, opened 2026-08-24
Two production sites read a TCP connect as "the external server is healthy"


BetterQAFTW:
## Business impact

Two code paths decide that an external ActivityWatch is alive and capturing on the strength of a bare TCP connect. A process that holds port 5600 but no longer answers HTTP satisfies that test, so on such a device the agent reports capture as healthy while nothing is being recorded — silent time loss, and the fleet is told everything is fine. This is the same conflation that produced the Critical fixed in PR 213, surviving at two addresses that PR deliberately did not touch.

## Steps to reproduce

Both were probed rather than read, varying only whether /info answers on a listener that accepts TCP:

    _start_component EBADARCH carve-out  (src/aw_manager.py:1851)
        -> tracker_download_failed = False   either way
    check_health external branch         (src/aw_manager.py:1325)
        -> True                              either way

## Expected result

A branch that concludes "an external server is capturing" asks the server, not the socket.

## Actual result

_port_in_use() is a bare TCP connect (a socket connect and nothing more). Both sites treat any listener as a healthy capturing server.

## Technical details

PR 213 fixed the third instance of this — the Rosetta preflight carve-out — by extracting _server_responding(), the /api/0/info ask that _wait_for_server already performed, and gating on it. That helper now exists and is the natural fix for both sites.

These two were left out of PR 213 deliberately, and the reasoning is worth carrying forward rather than re-deriving:

- Both are pre-existing and neither is reachable through the Rosetta preflight, which is what that PR was scoped to.
- Both are recoverable on a machine whose own binaries execute, because force_restart's reaper can kill a hung listener it owns. That is what makes the Rosetta case special: there, _reap_orphan_processes is path-scoped to binaries that have never run, so nothing can reap the listener and the only decision left is what to tell the person.
- The EBADARCH site sits under _lifecycle_lock on a per-watcher hot path, so adding a blocking HTTP call there needs thought about timeout and lock hold time rather than a copy of the fix.

**A note on why no test caught this.** _port_in_use is supplied by fixtures at six test sites, and at every one the surrounding prose calls it "external server healthy" / "capturing" / "running" — a claim a TCP connect cannot support. Every fixture supplies the boolean production computes, so no test in the suite can distinguish a serving ActivityWatch from a dead socket. Fixing these two sites without changing how they are tested would leave that gap intact; the fixtures need to express the port-answers-but-/info-does-not case, which none currently can.

## Environment

[...]

PR #217, opened 2026-08-24
Run the macOS trackers natively, so Rosetta 2 is no longer required

BetterQAFTW:
Bumps the ActivityWatch pin to **v0.14.0b4** and — the half that actually removes the requirement — teaches the preflight to ask what the **bundled binary** needs rather than what the host supports.

Handed over by the #213 session, which took #188's messaging half and explicitly did not take the bump.

## Why the version bump alone would have changed nothing

`_start_locked` gated every tracker start on `_rosetta_missing()`, which asks *"can this Mac execute x86_64?"*. Correct exactly while every bundled macOS tracker was x86_64 — true from the first release until this PR. Ship native arm64 binaries under that gate and a clean Apple Silicon Mac still answers "Rosetta missing" and still refuses to spawn binaries that need none. Rosetta stays mandatory, and no existing test says so.

The gate is now `_rosetta_required()`, the conjunction of two independent questions:

| question | answered by |
|---|---|
| does the **binary** need Rosetta? | `_bundled_trackers_need_rosetta()` — reads the Mach-O header of the tracker about to be spawned |
| does the **host** lack Rosetta? | `_rosetta_missing()` — unchanged |

The binary question is asked first because it is the cheap one. A native install now never forks `/usr/bin/arch` on the 60-second start path, and leaves `_rosetta_missing_cached` as `None`, so `capture_blocked_remedy()` correctly offers no Rosetta instruction on a machine that does not need one.

## Proof of failure

Same fixture, both gates, on a simulated Apple Silicon Mac with Rosetta absent:

```
NATIVE arm64 trackers (the #216 case)
  PRE-FIX  _rosetta_missing()  = True  -> BLOCKED
  POST-FIX _rosetta_required() = False -> allowed

x86_64 trackers (the control)
  PRE-FIX  _rosetta_missing()  = True  -> BLOCKED
  POST-FIX _rosetta_required() = True  -> BLOCKED
```

The control is the point: the new gate is not merely switched off. `test_x86_trackers_on_a_mac_with_no_rosetta_are_still_blocked` fails if it ever is, which would restore the 21-EBADARCH-failures-per-hour behaviour the preflight was built to end.

## The upgrade path, which the bump would otherwise have missed entirely

`_install_to_persistent` copies trackers to a stable path once, deliberately, so the Accessibility grant survives updates. That also means **an Apple Silicon machine that installed BetterFlow before this PR keeps its x86_64 trackers after updating to a native build** — and both copies are signed with our team, so the existing signing check sees nothing wrong. Those machines would have needed Rosetta forever while the release notes said otherwise.

`_should_reinstall_trackers` now also fires when the installed tracker cannot run natively here and the bundled one can. It is split into `_tracker_reinstall_reason`, returning a sentence rather than a bool, because the two reasons need different words — #213 exists because a surface named the wrong cause for 90 minutes, and one fixed log line for both repeats that one layer down.

## The pin had two definitions and nothing compared them

```
src/aw_manager.py       AW_VERSION + assets   <- verify_tracker_pins.py checked this
scripts/download_aw.py  a second copy         <- build.yml FETCHES with this
```

So the nightly guard verified digests for a version the build did not necessarily download. Both now import `src/aw_release.py`, the only place `AW_VERSION` is assigned, with a test asserting that and a positive control on its own regex.

Two related holes closed:

- **`verify-tracker-pins.yml` triggered on `src/aw_manager.py` only.** Moving the pin without updating its paths would have silently disarmed the nightly guard for every future pin edit.
- **`scripts/download_aw.py` verified no digest at all** — the build was the unverified half of a pair whose whole premise is that upstream assets are mutable under a pinned tag. It now fails closed, and re-downloads when the trackers on disk are the wrong architecture for the build (the per-OS output directory cannot distinguish them by name).

macOS gains an architecture dimension; Windows and Linux deliberately do not. Windows on ARM emulates x64 transparently, so keying it on arm64 would find no asset and break a configuration that works today. Upstream publishes no linux-arm64 at all, so that host keeps the x86_64 archive exactly as before — unchanged, not improved. A test pins the asymmetry so it is not "tidied" into symmetry.

## Verification

**Pins re-verified against upstream** rather than trusted from my own earlier measurement, using the repo's own instrument:

```
$ python scripts/verify_tracker_pins.py
[darwin-arm64]  OK 98a142c4...
[darwin-x86_64] OK 090b91b2...
[linux]         OK 5f608c7c...
[windows]       OK c7acb66d...
All 4 pins verified for AW_VERSION v0.14.0b4.   exit 0
```

**Real-environment smoke** — jsdom-equivalent unit tests cannot see any of this, so aw-server-rust 0.14.0b4 was actually run and driven with our own `AWClient`:

- every endpoint the client uses answers correctly (`buckets/`, `buckets/<id>`, `buckets/<id>/events`, `buckets/<id>/heartbeat`), buckets parse, events read back
- **schema upgrade**: 0.13.2 leaves the DB at v4, 0.14 migrates it to v5. A v4 database written by the old server migrated with **all 3 events intact**, and accepted new writes afterwards
- **rollback**: the old 0.13.2 server still reads the upgraded v5 database and sees all 4 events, so a downgrade does not strand or lose data

Binaries were verified as `Mach-O 64-bit executable arm64` and run natively (`--help`, rc=0) on an M-series Mac; every CLI flag we pass (`--timeout`, `--poll-time`, `--strategy {jxa,applescript,swift}`, `--port`, `--dbpath`) is present and unchanged.

**Suite and lint** (CI does not run in this org, so both were run locally, replicating the `test` job):

```
pytest tests/ -q     1914 passed, 1 skipped
ruff check src/ tests/   180 errors  (identical to origin/main — zero added)
```

## Not in scope

- **#215** (the two remaining bare-TCP health reads) is untouched. Separate issue, the #213 session's finding.
- The 0.14 line is a **prerelease**. The last stable, v0.13.2, is from Oct 2024 and publishes no arm64 macOS build at all, so there is no non-beta route off Rosetta. That tradeoff is the main thing worth a human opinion before merge.

Refs #216, #188.


BetterQAFTW:
## Review round: 4 findings fixed, CI now green

Two adversarial lenses ran on this branch. **CI is green as of `a60f078`** (it was red for the whole review window, which is the first thing below).

### Fixed

**1. CI was red on Linux and my local number was not platform-neutral.** `_rosetta_required()` is macOS-only and short-circuits on `sys.platform` *before* consulting either half of its conjunction. Seven tests patched `_rosetta_missing` — which used to BE the gate — so they blocked nothing on the runner and `_start_locked` sailed past and spawned. An eighth counted `/usr/bin/arch` forks with the binary half left to the real resolver, so it was measuring *whatever trackers happen to be installed on the machine running the suite*: x86_64 on this laptop (green), ELF on the runner (red).

Worth stating plainly: my earlier "1914 passed" was green partly because this laptop has BetterFlow installed. The runner's 11 skips against my 1 was the tell, and I did not read it.

**2. `make ship` put one architecture's trackers into BOTH macOS DMGs.** `ship-arm64` / `ship-x86_64` call PyInstaller directly with no `download-aw` prerequisite. Harmless while macOS had a single asset — correct by construction — and a live defect the moment this branch added the second. On a real Intel Mac the runtime gate cannot catch it: `_rosetta_required()` → False (host is not arm64), `_tracker_reinstall_reason()` → None (the bundled copy is equally unrunnable there). It would spawn and EBADARCH every 60s forever: **#216 recreated on Intel by the fix for #216.**

Each leg now fetches its own arch, `download-aw` exports `TARGET_ARCH` (`TARGET_ARCH ?=` is a make variable, so the script never saw it), and **`build.spec` now fails the build** rather than package a mismatch — the only check on that path, since `make ship` bypasses `download_aw.py`'s own guard. Verified in all four directions against real binaries.

**3. The upgrade path could not repair a LEGACY FLAT install** — the population this PR exists to rescue. `_install_to_persistent` removed the destination only when it was a directory; a flat install puts a *file* there, so `copytree` raised `[Errno 17] File exists` and the reason, recomputed every cycle, stayed identical forever. Removing the launcher is also what fixes resolution, since `_resolve_binary_path` checks the flat path first. The caller discarded the return value, so it was invisible in the log too.

**This was Phantom 7 and it had full green coverage.** Every reinstall test asserted the *reason* — the producer — and none asserted the tree's architecture actually changed. Two consumer tests now do, and reverting only the new branch reproduces the exact `[Errno 17]` and reddens exactly the flat test.

**4. `TARGET_ARCH` was not in the download step's environment in CI**, so the protection `get_asset_key()` exists to provide was inert; and `verify-tracker-pins.yml` did not trigger on `src/machine_arch.py`, which `aw_release.py` now imports for the spellings the asset keys are built from — the same trigger-path hole this branch already closed for `aw_manager.py`.

### Measured, not fixed

- **A mixed-architecture payload would pass the arch gate**, which resolves only the three launchers while the watchers ship an embedded `Python.framework` and `lib-dynload/*.so`. Measured on the real v0.14.0b4 arm64 archive: **166 Mach-O files across the three components, zero x86_64.** Not present in this pin. The design point stands as a latent risk for a future one.
- **WAL is the part a rollback does not undo.** `aw-server-rust` sets `journal_mode=WAL`, which is written into the file header and persists after the process exits, including for a rolled-back 0.13.2 that never asked for it. Probed read-only and it does not break, so this is backward-compatible — but it is the irreversible half, and the v4→v5 migration (index-only, `BEGIN EXCLUSIVE`, no table alteration) is not.

### Still open, and not mine to decide

- **No version stamp.** Nothing records which AW version or DB schema a machine is on — not on disk, not on the heartbeat. After merge there is no way to answer "has this machine migrated?", which is precisely the data a rollback decision needs.
- **No kill switch.** The AW version is a compile-time constant; no server config can change it. The `beta` channel exists but `update_channel` is settable only in the tray, per machine, and is not in `HEARTBEAT_HEALTH_KEYS`, so a pilot group cannot even be enumerated.

Both point the same way: if this ships as a pilot rather than fleet-wide, the version stamp should probably be part of this PR. Flagging for a human rather than deciding it.

BetterQAFTW:
## Rebased onto main, 2026-08-26. One conflict, and it was the version.

The branch had gone `CONFLICTING/DIRTY` because #222 landed on `main` on 2026-08-25, in the same Rosetta code this branch rewrites.

**The conflict was `src/__init__.py`, and neither side was right.** The branch bumped to `1.5.126`; `main` has since shipped `1.5.127`. Taking the branch's would release backwards, taking main's would ship this change unversioned. Resolved to **1.5.128**. `src/__init__.py` is the only live carrier — the other `1.5.12x` hits in the repo are docs and workflow comments naming historical versions.

**`aw_manager.py` auto-merged, so I checked the result rather than trusting it.** #222's carry-through survives — `_rosetta_missing_conclusive` at 5 references plus its `keeping last answer` log line — and `tray.py` still derives its label through `process_arch`. Both landed after this branch was cut, and both live in the code this branch restructures.

### Gate

```
pytest tests/          2018 passed, 1 skipped, exit 0
```

That is up from main's ~1823 baseline, which is this branch's own new coverage (`test_native_arm64_trackers.py` alone is 532 lines).

**Two local results that are NOT findings, said out loud rather than omitted:**

`scripts/verify_tracker_pins.py` fails here with `read operation timed out` and `nodename nor servname provided` — it downloads the pinned tracker artifacts, and this machine could not reach them. Not a pin mismatch. CI's `verify-pins` job passes, which is the run with network, and `test_aw_pin_single_source.py` covers the consistency half without it.

`ruff check .` reports 187 errors, 9 of them in files this branch touches. **Ruff runs in no workflow in this repo** — it is not a gate, and this is ungated repo-wide lint debt rather than anything this change introduced.

Pushed as `1.5.128`; CI re-runs against the merged tree.

PR #218, opened 2026-08-24
Say which kind of 'not right now' this is, on all eight states that know

### LINKED ISSUE #214: The tray hides the specific reason on five states, showing a generic label instead
## Business impact

When the agent stops recording, the tray's persistent line tells the user *what kind* of state it is in, but not *why*. The code already works out the specific reason and stores it — then throws it away and shows a fixed label. So a person whose session expired, a person whose queue filled up, and a person outside working hours are all shown the same unhelpful word, and the one sentence that would tell them what to do never reaches them.

The agent's own source says why this matters, at `src/main.py:2846`: "Say which kind of 'not right now' this is. A keychain we cannot open is not an offline agent, and telling the user it is sends them to check their wifi." That sentence is written, stored, and discarded.

PR #213 fixed exactly this defect for the ERROR state, because that is the state issue #188 was about. The same defect is still live on five others.

## Steps to reproduce

1. Run the agent and put it into any of the affected states — the easiest is to let the sync queue fill (QUEUE_WARNING), or sign out so the session needs re-auth (WAITING_AUTH).
2. Open the tray menu and read the "App status:" line.

## Expected result

The line reads the specific sentence the agent already computed, e.g. `Queue 87% full`, or `Session expired, re-login required`.

## Actual result

The line reads a fixed generic label — `Offline (queue full)`, `Waiting for login...`, `Paused`, `Outside working hours`, `Offline` — the same text regardless of which underlying cause produced the state.

## Technical details

Verified by reading `origin/fix/mac-arch-detection-188-184` at `0b7db805` on 2026-08-24.

`TrayIcon._get_status_text` (`src/ui/tray.py`) returns a hard-coded string for every state except ERROR, which PR #213 changed to render `model.status_text`. The writers whose text is still discarded:

| writer | state | text written |
|---|---|---|
| `src/main.py:1593` -> `:1876` | QUEUE_WARNING | `f"Queue {capacity_pct}% full"` |
| `src/main.py:1589-1591` -> `:1876` | PRIVATE_HOURS | `"Private hours — not recording (draining N queued)..."` |
| `src/main.py:2846-2848` | QUEUED | `"Reading saved sign-in..."` / `"Offline — reconnecting..."` |
| `src/main.py:2205`, `:2851`, `:2900`, `:3411` | WAITING_AUTH | `"Session expired, re-login required"` |
| `src/system_event_handler.py:103`, `:170`; `src/idle_manager.py:383` | PAUSED | `"Sleeping"` / `"Screen locked"` / `"Idle"` |

Two structural notes for whoever picks this up:

- `set_state` defines `faulted = (TrayState.ERROR, TrayState.QUEUE_WARNING)` and applies both the entry-clear and the recovery-clear to that pair, while `_get_status_text` renders the field for ERROR only. One rule — "which states own `status_text`?" — is written in two places with two different answers, and QUEUE_WARNING is where they disagree.
- `set_state`'s docstring still says "Optional status message for error state" while eight call sites pass one for other states.

Note the ordering hazard: rendering `status_text` for a state also requires the entry/recovery clears to cover that state, or a stale sentence from a previous state leaks under the new label. PR #213 added the entry-clear for exactly this reason.

## Environment

All platforms (macOS, Windows, Linux) — this is platform-independent tray logic. Branch `fix/mac-arch-detection-188-184` at `0b7db805`; also present on `main` at `4a1f2ab`. Raised by the pre-merge review of PR #213.

tudorsss:
`_get_status_text` rendered `model.status_text` for ERROR and threw it away everywhere else. Seven states composed a specific sentence, stored it, snapshotted it, and showed the user one generic word instead.

Follows PR #213, which made ERROR read the field because ERROR is the state #188 was about.

## Proof of failure

Captured against the pre-fix tree before any edit — the real writers' states with the real text they write:

```
state              written                               rendered
queue_warning      Queue 87% full                        'Offline (queue full)'   DISCARDED
private_hours      Private hours - not recording (...)   'Outside working hours'  DISCARDED
queued             Offline - reconnecting...             'Offline'                DISCARDED
waiting_auth       Session expired, re-login required    'Waiting for login...'   DISCARDED
paused             Screen locked                         'Paused'                 DISCARDED
needs_permissions  Input Monitoring permission required  'Permissions needed'     DISCARDED
starting           Restoring your session...             'Starting...'            DISCARDED
error              the #213 control                      'the #213 control'       OK
```

The ERROR row is the control: it proves the probe can see a rendered sentence, so the seven DISCARDED rows are the subject failing rather than the instrument.

**Eight owners, not the five the issue listed.** `NEEDS_PERMISSIONS` writes its text directly under the model lock (`src/main.py:810`) rather than through `set_state`, and `STARTING` receives one from `_set_startup_status`. Both are invisible to a grep for `set_state(TrayState.X, "…")`, which is how the issue's table came to be short by two.

## The defect is the membership question, not the eight branches

`set_state` decided which states own the field — `faulted = (ERROR, QUEUE_WARNING)` — and `_get_status_text` decided again, differently, for ERROR alone. One rule, two implementations, disagreeing at QUEUE_WARNING.

That is not cosmetic. The clears are the only thing stopping one state's sentence rendering under the next state's label, so *"renders it"* and *"is cleared on entry to it"* have to name the same set. A branch widened without its clear is worse than the generic label it replaced.

- `STATUS_TEXT_STATES` maps each owning state to its floor. Membership and fallback live together, so a state cannot join the render path without a label and show `App status:` with nothing after it.
- `set_state` reads that dict for both clears. The stale comment claiming the set "holds TWO states" is corrected — widening it takes that hole from one ordered pair to fifty-six.
- `_owned_status_text` applies ERROR's existing floor to all eight, and uses `s.get()` so a caller building a partial snapshot by hand still gets a label rather than a `KeyError`.
- `STARTING` is an explicit member rather than folded into the `else` fallback, which stays hard-coded — a fallback that rendered the field would hand an unknown state a stale sentence.

## One existing test asserted the opposite, deliberately inverted

`test_transition_between_two_healthy_states_preserves_existing_text` pinned SYNCING → PAUSED *preserving* a leftover sentence, justified in its own docstring by the writer possibly being "a permissions hint". That was safe only while PAUSED rendered a constant and the field was never shown. Now it would put a permissions hint under a Paused icon on a sleeping laptop.

Replaced with the clear, plus a control (SYNCING → PRIVATE, neither state owning the field) that keeps the half still true — so a clear-everything fix cannot pass by discarding state nothing was harming.

## Mutation matrix

Five mechanisms, each reddening **distinct** named tests. Control clean before and after; tree restored to 0 dirty files.

| mutant | witnesses |
|---|---|
| M1 render → generic label | `test_the_error_status_carries_the_sentence_it_was_given`, `test_the_menu_row_the_user_reads_carries_it`, `test_every_state_either_owns_the_field_or_provably_ignores_it` (+3) |
| M2 entry-clear removed | `test_a_sentence_never_survives_into_a_different_state`, `test_entering_a_state_that_renders_the_field_clears_a_foreign_sentence` (+4) |
| M3 recovery-clear removed | `test_recovery_from_error_clears_stale_status_text`, `test_recovery_from_queue_warning_clears_stale_status_text` |
| M4 floor removed | `test_a_missing_or_junk_sentence_falls_back_to_the_states_own_label`, `test_blank_text_falls_back_rather_than_rendering_an_empty_row` |
| M5 read leaks to every state | `test_every_state_either_owns_the_field_or_provably_ignores_it`, `test_the_other_states_are_untouched_by_the_error_branch` (+4) |

M1 and M2 were separated explicitly (`comm` on their reddened sets) because they share one test — each has six witnesses the other does not, so neither mechanism is riding on the other's coverage.

Two instrument bugs were caught and are worth recording: the first matrix printed **nothing at all** because the helper returned `cmp`'s status, so "files differ" (mutation applied) read as failure and the test run never fired — that would have looked like a clean matrix. And the first version of the render test asserted `generic not in row`, which cannot pass against a correct fix, because QUEUED's real sentence is `Offline — reconnecting...` and contains its own generic label.

The membership guard walks every `TrayState` member and asserts the partition in both directions, so a state added later is covered without anyone extending a list here.

## Verification

- `pytest tests/` → **1899 passed, 1 skipped, exit 0**. That is the whole of what CI's `test` job runs (`pytest tests/ -v --tb=short`, ubuntu-latest); no other non-deploy workflow touches this code.
- `ruff check src/ui/tray.py` → 7 errors, all pre-existing at lines 37–121, rule-set byte-identical to `origin/main`'s. Baseline taken from `git show origin/main:src/ui/tray.py` rather than from the working tree.
- Platform-independent tray logic; no real-environment surface, so no browser/API smoke applies.

Closes #214.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -814,6 +814,16 @@ class SyncCoordinator:
                 else:
                     if current_state == TrayState.NEEDS_PERMISSIONS:
                         self.tray.model.state = TrayState.SYNCING
+                        # Clear the hint we wrote above. This block is the only
+                        # place that moves state WITHOUT going through
+                        # set_state, so neither of its clears runs here. The
+                        # stale text is currently unreachable — SYNCING does not
+                        # render the field, and every later entry into a state
+                        # that does trips the entry-clear — but that is an
+                        # argument about the whole state machine holding, made
+                        # afresh by anyone who adds an owner. One line removes
+                        # the dependency on it.
+                        self.tray.model.status_text = None
                         should_update_icon = True
                     else:
                         should_update_icon = False
@@ -1808,7 +1818,8 @@ class SyncCoordinator:
             is_idle = self.idle_paused
 
             if self.paused_by_network:
-                self.tray.set_state(TrayState.QUEUED, "Offline")
+                # No text: the label is STATUS_TEXT_STATES[QUEUED].
+                self.tray.set_state(TrayState.QUEUED)
                 self.tray.update_stats(queue_size=self.queue.size())
                 # A network outage suspends UPLOAD, not capture: suspend_upload()
                 # keeps the trackers recording and queueing locally, often for
@@ -2022,7 +2033,8 @@ class SyncCoordinator:
         except Exception:
             reachable = False
         if not reachable:
-            self.tray.set_state(TrayState.QUEUED, "Offline")
+            # No text: the label is STATUS_TEXT_STATES[QUEUED].
+            self.tray.set_state(TrayState.QUEUED)
         else:
             self.tray.set_state(TrayState.ERROR, error_message)
 
--- a/src/system_event_handler.py
+++ b/src/system_event_handler.py
@@ -144,6 +144,11 @@ class SystemEventHandler:
             return
         if self.coordinator.is_on_break:
             logger.info("System wake - staying on break")
+            # Second reason to stay paused, same stale sentence as the branch
+            # above. PAUSED is re-asserted rather than ON_BREAK because the
+            # state is not this handler's to change — it only has to stop the
+            # tray claiming a cause that has ended.
+            self.tray.set_state(TrayState.PAUSED)
             return
         # Private Time is intentionally NOT auto-restored: it was ended at the
         # sleep boundary (on_system_sleep), so a sleep cleanly ends a private
@@ -203,6 +208,11 @@ class SystemEventHandler:
             return
         if self.coordinator.is_on_break:
             logger.info("Screen unlocked - staying on break")
+            # Second reason to stay paused, same stale sentence as the branch
+            # above. PAUSED is re-asserted rather than ON_BREAK because the
+            # state is not this handler's to change — it only has to stop the
+            # tray claiming a cause that has ended.
+            self.tray.set_state(TrayState.PAUSED)
             return
         if pre_lock_private:
             logger.info("Screen unlocked - restoring private time")
--- a/src/system_event_handler.py
+++ b/src/system_event_handler.py
@@ -134,6 +134,13 @@ class SystemEventHandler:
                 logger.warning("send_sleep_event failed: %s", e)
         if user_paused:
             logger.info("System wake - staying paused (user-initiated pause active)")
+            # Re-assert PAUSED with no text. The state is unchanged, so this is
+            # a no-op for the icon — but the tray now RENDERS status_text, and
+            # the sentence still sitting there is "Sleeping", written on the way
+            # down. Returning without this leaves an awake laptop reporting that
+            # it is asleep. The cause is now the user's pause, which has no
+            # sentence of its own: the generic "Paused" is the true answer.
+            self.tray.set_state(TrayState.PAUSED)
             return
         if self.coordinator.is_on_break:
             logger.info("System wake - staying on break")
@@ -189,6 +196,10 @@ class SystemEventHandler:
             pre_lock_private = self._pre_lock_private
         if user_paused:
             logger.info("Screen unlocked - staying paused (user-initiated pause active)")
+            # Same as the wake path above: "Screen locked" is still stored and
+            # now renders, so an unlocked screen would keep reporting itself
+            # locked.
+            self.tray.set_state(TrayState.PAUSED)
             return
         if self.coordinator.is_on_break:
             logger.info("Screen unlocked - staying on break")
@@ -237,4 +248,6 @@ class SystemEventHandler:
             logger.info("Network offline - suspending upload (capture continues)")
             self.sync_engine.suspend_upload("network offline")
             self.coordinator.paused_by_network = True
-            self.tray.set_state(TrayState.QUEUED, "Offline")
+            # No text: "Offline" is STATUS_TEXT_STATES[QUEUED] and passing it
+            # here would be a second copy of the label to keep in step.
+            self.tray.set_state(TrayState.QUEUED)
--- a/src/ui/tray.py
+++ b/src/ui/tray.py
@@ -449,6 +449,42 @@ class TrayState(Enum):
     WAITING_AUTH = "waiting_auth"  # Amber - waiting for browser login
 
 
+# The one answer to "which states own ``status_text``?", and the label each one
+# falls back to when its writer had nothing specific to say.
+#
+# It was previously two answers. ``set_state`` cleared the field for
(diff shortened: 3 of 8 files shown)


tudorsss:
## Review rounds 2 and 3 — what changed since the PR body was written

**Round 2 (two independent lenses, both on the round-1 diff): both found the same regression, which I had introduced.**

`set_state` skips the entry-clear when the state does not change. That retention is deliberate for ERROR — "the same outage escalating again" — and was safe while the only owners were two single-cause fault states. Widening the owning set extended it to PAUSED, which has five writers with five different causes:

```
idle pause, then user clicks Pause  ->  'Idle'           (want 'Paused')
lock, then unlock while user-paused ->  'Screen locked'  (want 'Paused')
sleep, then wake while user-paused  ->  'Sleeping'       (want 'Paused')
```

All three read `Paused` on `origin/main`. The pause case fires `send_notification("Tracking Paused")` in the same call, so the notification and the tray row contradicted each other; the sleep case told a user at an awake laptop that it was asleep.

Retention is now data (`STATUS_TEXT_RETAIN_ON_REENTRY`) — a state earns it by having **one** cause, not by owning the field. The two lenses disagreed on membership (`{ERROR}` vs `{ERROR, QUEUE_WARNING}`) and the suite settled it: QUEUE_WARNING's retention is pinned by an existing test, and it has exactly one writer, which always passes text.

**Three of my own tests were proven worthless, by mutation rather than by argument.**

- The `inspect.getsource` guard was wrong in *both* directions: restoring the exact defect left it **green** (the literal it greps for also appears in the comment above the line), and a behaviour-preserving extraction turned it **red**. Deleted; replaced by a walk over all 121 ordered state pairs, which asserts the rule rather than the text.
- The STARTING fallback check read back the model **seed**, which happens to equal the label — a mutant of that dict entry survived the whole module.
- The partial-snapshot test could not distinguish "does not read the key" from "reads a key I did not supply". That is the same blindness that let #213's own guard miss this widening.

**Round 3: the refutation reviewer died on an account spend limit before reading the diff.** Its attack list was run by hand instead. Saying so plainly, because a self-administered refutation on one's own code is the weaker instrument — and it is the reason the last independent look was at `00e5a5e`, three commits back.

It found one real defect anyway, and it is the class round 2 half-closed. Both resume handlers have **two** reasons to stay paused; round 2 fixed the manual-pause branch because that was the instance under the microscope, and the break branch two lines below had the identical stale sentence. Anyone whose laptop sleeps or locks during a break saw it.

The class is now swept rather than the instance patched: all four early returns re-assert, and every other PAUSED writer either passes text or lands somewhere that clears.

Also checked and clean: lock ordering for the two new `set_state` calls (both sit after their `_pause_state_lock` block closes, so `model.lock` is never taken while it is held); QUEUED can never show a foreign sentence (not in the retain set, so the entry-clear fires on every text-less entry); and `main.py`'s NEEDS_PERMISSIONS null-out sits inside the `granted == True` branch, so it cannot destroy a hint that is still needed.

**Mutation matrix, round 3.** Two of the instruments were dead before they were trusted: the harness printed nothing because it returned `cmp`'s status, then reported `no tests ran` five times because zsh does not word-split an unquoted variable — every mutant would have read as "survived". Once fixed it showed the wake/unlock fix had **no witness at all**; reverting both re-asserts left all 49 tray tests green. Now each path reddens its own case.

Suite: **1905 passed, 1 skipped, exit 0**.

Issue #211, opened 2026-08-23
Failed self-update strands a device on an old backup copy, and the leftover duplicate app copies break the Accessibility grant


BetterQAFTW:
## Business impact

The employee's dashboard shows "Window titles empty" while their System Settings show Accessibility granted, so the badge reads as a false alarm and nobody can act on it. Meanwhile the device silently runs an old agent version (below the server's own minimum-version floor) for days, because the self-update fails and the machine keeps relaunching the stale backup copy. Window-title categorization data is lost for the whole period, and time review/billing accuracy degrades with it.

## Steps to reproduce

Observed on a real device (Martin's Mac, device 14, `sync:456bc69b-001`), 2026-08-23:

1. Let a self-update fail mid-apply (here: `hdiutil detach failed (rc=1)` aborting the 1.5.119 → 1.5.125 apply).
2. Note `/Applications` now holds three copies with the SAME bundle id `co.betterqa.betterflow`: `BetterFlow.app` (1.5.125), `BetterFlow.app.old` (1.5.120), `BetterFlow-1.5.119-backup.app` (1.5.119).
3. Reboot/relogin — the machine comes up running the 1.5.119 backup copy (`self_updater` log: `sys.executable=/Applications/BetterFlow-1.5.119-backup.app/Contents/MacOS/BetterFlow`), even though `~/Library/LaunchAgents/co.betterqa.betterflow.plist` says `open -a /Applications/BetterFlow.app`.
4. Observe the admin dashboard "Window titles empty" badge on the device, with the Accessibility toggle ON in System Settings.

## Expected result

A failed update retries or rolls back cleanly; the device never keeps running a version below the server's minimum floor for days; old app copies are cleaned up once a newer version is confirmed running; the Accessibility grant, once given, keeps working.

## Actual result

The device ran 1.5.119 until 14:54 today while the server logged `Agent 1.5.119 is below minimum 1.5.124 — update required` on every sync. Accessibility trust FLAPPED, including two mid-run revocations with no restart and no update in between:

```
2026-08-22 09:08:00 GRANTED
2026-08-23 10:49:25 DENIED   <- revoked mid-run
2026-08-23 13:54:58 GRANTED  <- user re-toggled
2026-08-23 14:26:25 DENIED   <- revoked mid-run again, ~30 min later
2026-08-23 14:51:12 GRANTED
[...]

PR #219, opened 2026-08-24
Stop the duplicate app copies picking which agent version runs

tudorsss:
Two of the four mechanisms in #211, the pair that together remove the duplicate-copy hazard. **Does not close the issue** — the other two are listed at the bottom.

## What happened on the device

`/Applications` held three copies — `BetterFlow.app` (1.5.125), `BetterFlow.app.old` (1.5.120), `BetterFlow-1.5.119-backup.app` (1.5.119) — all signed by us, all registered under `co.betterqa.betterflow`. The machine booted the **oldest** and sat below the server's own minimum version floor for days while every sync logged `update required`. The Accessibility grant flapped, twice revoked mid-run with no restart in between, because macOS keeps one row per app and which copy that row is understood to mean is not something we control.

## 1. `open -a <path>` was letting LaunchServices choose

`_app_launch_args` built `open -a /Applications/BetterFlow.app`. The two forms look interchangeable and are not: `-a` names an **application**, resolved through the LaunchServices database, so with several copies under one bundle id the choice is LaunchServices' and not ours. Dropping `-a` makes the argument a path again.

```
assert "-a" not in ["open", "-a", "/Applications/BetterFlow.app"]   FAILED pre-fix
assert args == ["open", "/Applications/BetterFlow.app"]             FAILED pre-fix
assert _macos_plist_is_current() is False                           FAILED pre-fix
```

The non-bundled control passed pre-fix and still passes, so the tests isolate the defect rather than describing the new code.

**Scope, stated honestly:** the issue marks the LaunchServices resolution as INFERRED — tccd logs had aged out of unified-log retention, so nobody watched macOS pick the wrong copy. This is not "the proven fix for that morning". It removes the only degree of freedom the launch command had, which is worth doing whether or not it was the cause.

Migration needs no code and is **asserted rather than assumed**: `_macos_plist_is_current()` compares installed ProgramArguments against this list, so every plist in the fleet still carrying `-a` now reads as stale and gets rewritten on next startup. If that assertion ever flips, the fix silently reaches nobody.

## 2. The copies themselves

Neither leftover name is one current code produces — `_apply_macos_update` makes `<stem>.old.app` and removes it on success; `.app.old` is the Linux AppImage shape. They are sediment from older versions, which is exactly why a **sweep** is needed rather than a fix to whichever writer is current: the copies already exist in the fleet and no future correctness removes them.

This deletes directories in `/Applications`, so most of the 15 tests are about what it must **refuse** to touch. Four guards:

| | guard | why |
|---|---|---|
| 1 | **Identity** — the candidate's own `Info.plist` must name our bundle id | The one doing real work. A name pattern is a guess about provenance; an `Info.plist` is the bundle stating who it is. Unreadable means "I don't know whose this is", never "delete it". |
| 2 | **Name** — closed list of three shapes we can show we produced | Not a glob. A leftover costs disk; a wrong deletion costs somebody an application. |
| 3 | **Siblings only**, never recursive | A directory walk of `/Applications` is not something this should grow into. |
| 4 | **No symlinks** | Deleting through one deletes its target, which by definition is not a copy our updater left. |

`find_stale_bundle_copies` returns paths and deletes nothing, so the decision is testable without a filesystem that can lose something. `purge_` does the removal, best-effort — failing to tidy up must never stop the agent starting.

New behaviour, so no pre-fix failure exists to capture. The tests were earned with mutants instead:

```
identity guard removed        -> 2 named tests red
unreadable Info.plist = ours  -> 1 named test red
symlinks become candidates    -> 1 named test red
the startup call deleted      -> 1 named test red
```

### One mutant survives, deliberately

`entry == running_app` is unreachable while the name patterns are correct — a bundle's own name cannot match a pattern built from its own stem. That is defence in depth, not a gap, and it was proven as a **pair** rather than argued:

```
neuter the name filter alone     -> running bundle NOT returned
neuter the name filter AND this  -> running bundle IS returned
```

So it is the only thing standing between a broken name pattern and deleting the app that is currently executing. Recorded at both the call site and the test, so the next mutation run doesn't read a surviving mutant as dead code and remove the last guard on the running app.

### Wiring

At startup, not on the post-update path — the copies that matter are already on disk, and fixing the writer does nothing for a machine that already has them. Placed after `_maybe_relocate_to_applications()`, which can move us, since the sweep is scoped to siblings of where we are *now*.

A helper nobody calls would leave the device exactly as it was, so the callsite has its own guard: it matches on a **call** and strips comment lines first, with a control proving the stripping happens, so it cannot be satisfied by the prose explaining it.

`BUNDLE_ID` was already spelled in `autostart.py` (as the launchd label) and `ui/permissions.py`. The third copy is pinned by a test rather than refactored — threading one constant through the tray and the launch agent couples modules with no other reason to know about each other, for a string that changes approximately never.

## Verification

- `pytest tests/` → **1924 passed, 1 skipped, exit 0**. That is the whole of CI's `test` job.
- `ruff` → `src/self_updater.py` back to *All checks passed*, 8 errors total across the three touched files, identical to `origin/main`. An earlier revision introduced two `E402`s and an `I001` by inserting the constant block between imports — caught by diffing against baseline, not by reading.
- `git diff --numstat` → one deleted line in the whole branch (the `-a` argument).
- **Not smoke-tested on a real Mac with real duplicate copies.** The sweep's filesystem behaviour is covered by tmp_path fixtures; LaunchServices resolution cannot be tested at all off-device. Both are macOS-only paths that the ubuntu CI runner never executes.

## Still open on #211

- Recover from a failed `hdiutil detach` (or copy out of the mount rather than requiring detach before install) — the actual trigger of the incident.
- Alert ops when a device has been below the server's minimum version floor for more than a day.

Refs #211.

### DIFF
--- a/src/autostart.py
+++ b/src/autostart.py
@@ -114,7 +114,29 @@ def _app_launch_args() -> list[str]:
         # Use 'open -a' with the .app bundle path for a clean launch
         parts = exe.split(".app/Contents/MacOS/")
         bundle_path = parts[0] + ".app"
-        return ["open", "-a", bundle_path]
+        # `open <path>`, NOT `open -a <path>`. They look interchangeable and are
+        # not: `-a` names an APPLICATION, which macOS resolves through the
+        # LaunchServices database rather than by opening the path you handed it.
+        # When several copies are registered under one bundle id, which copy
+        # wins is LaunchServices' choice.
+        #
+        # #211: a device's plist said `/Applications/BetterFlow.app` and the
+        # machine came up running `/Applications/BetterFlow-1.5.119-backup.app`
+        # — below the server's minimum floor — for days, while every sync logged
+        # "update required". Three copies shared `co.betterqa.betterflow`.
+        # Without `-a` the argument is a path again and no other bundle can be
+        # substituted.
+        #
+        # The resolution itself is INFERRED, not witnessed (the issue says so;
+        # tccd logs had aged out). This does not depend on that being the cause:
+        # it removes the only freedom the launch command had, which is worth
+        # doing either way.
+        #
+        # Migration is automatic and needs no code. `_macos_plist_is_current()`
+        # compares the installed ProgramArguments against this list, so every
+        # plist still carrying the `-a` form now reads as stale and
+        # `ensure_synced()` rewrites it on the next startup.
+        return ["open", bundle_path]
     # Running as a Python script
     return [sys.executable, "-m", "src.main"]
 
--- a/src/main.py
+++ b/src/main.py
@@ -2980,6 +2980,33 @@ class BetterFlowApp:
             except Exception as e:
                 logger.warning("Auto-start sync failed (non-fatal): %s", e)
 
+        # Sweep sibling copies of ourselves left by earlier updates (#211).
+        #
+        # Runs at startup rather than after an update, because the copies that
+        # matter are ALREADY on disk — a device carried three, booted the oldest
+        # (below the server's minimum version floor) for days, and had its
+        # Accessibility grant flap, since macOS keeps one row per app and which
+        # copy that row means is not ours to decide. Fixing whichever writer
+        # leaves them behind does nothing for a machine that already has them.
+        #
+        # After _maybe_relocate_to_applications() above, deliberately: that call
+        # can move us, and the sweep is scoped to siblings of where we are NOW.
+        #
+        # Non-fatal on every path. Failing to tidy up must not stop the agent
+        # starting, which is the one job it has.
+        if sys.platform == "darwin" and ".app/Contents/MacOS/" in sys.executable:
+            try:
+                try:
+                    from .self_updater import purge_stale_bundle_copies
+                except ImportError:
+                    from self_updater import purge_stale_bundle_copies
+                running_bundle = Path(
+                    sys.executable.split(".app/Contents/MacOS/")[0] + ".app"
+                )
+                purge_stale_bundle_copies(running_bundle)
+            except Exception as e:
+                logger.warning("Stale-copy sweep failed (non-fatal): %s", e)
+
         # First-run setup wizard.
         #
         # Gate on stored CREDENTIALS, not on the setup_complete flag alone. The
--- a/src/main.py
+++ b/src/main.py
@@ -2980,32 +2980,7 @@ class BetterFlowApp:
             except Exception as e:
                 logger.warning("Auto-start sync failed (non-fatal): %s", e)
 
-        # Sweep sibling copies of ourselves left by earlier updates (#211).
-        #
-        # Runs at startup rather than after an update, because the copies that
-        # matter are ALREADY on disk — a device carried three, booted the oldest
-        # (below the server's minimum version floor) for days, and had its
-        # Accessibility grant flap, since macOS keeps one row per app and which
-        # copy that row means is not ours to decide. Fixing whichever writer
-        # leaves them behind does nothing for a machine that already has them.
-        #
-        # After _maybe_relocate_to_applications() above, deliberately: that call
-        # can move us, and the sweep is scoped to siblings of where we are NOW.
-        #
-        # Non-fatal on every path. Failing to tidy up must not stop the agent
-        # starting, which is the one job it has.
-        if sys.platform == "darwin" and ".app/Contents/MacOS/" in sys.executable:
-            try:
-                try:
-                    from .self_updater import purge_stale_bundle_copies
-                except ImportError:
-                    from self_updater import purge_stale_bundle_copies
-                running_bundle = Path(
-                    sys.executable.split(".app/Contents/MacOS/")[0] + ".app"
-                )
-                purge_stale_bundle_copies(running_bundle)
-            except Exception as e:
-                logger.warning("Stale-copy sweep failed (non-fatal): %s", e)
+        self._sweep_stale_bundle_copies()
 
         # First-run setup wizard.
         #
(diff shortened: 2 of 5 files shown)


tudorsss:
## Review rounds — two lenses, nine Important findings, all fixed

### The headline: the sweep did nothing on the exact device #211 reports

Both reviewers found it independently. Every pattern is built from the **running** bundle's stem, so from a non-canonical copy it cannot see the canonical install's siblings. Measured against the incident's own `/Applications` layout:

```
running canonical BetterFlow.app  -> finds both duplicates
running the 1.5.119 backup copy   -> finds []      <- what #211's device booted
running BetterFlow.app.old        -> finds []
```

Returning `[]` is still the right action — sweeping from a backup would have it delete the *newer* app beside it. The defect was that it happened **silently**. It now warns, naming the non-canonical bundle, because #211 was found by grepping one machine's log and that line is the only mechanism by which the fleet's remaining cases surface.

Not a regression: `_app_launch_args` already derived from `sys.executable` under the old form, so the pinning predates this branch. Recorded as a known limitation rather than a claim that the issue is closed.

### A new test would have killed the next release

`tests/test_stale_app_copy_cleanup.py` carried the **first `symlink_to` in the whole suite**. The tag build runs `pytest tests/` on the four-platform matrix including `windows-latest`, and `release: needs: build` — so a hard failure there produces **no artifacts at all**, and it would surface as a mystery failed release rather than a red PR, because the PR leg is ubuntu-only. Exactly how `os.getuid()` killed v1.5.119's first tag build.

Verified by simulating the refusal rather than reasoning about it:

```
OSError(1314, "A required privilege is not held by the client")
-> 27 passed, 1 skipped     (was: 1 failed)
```

### The half that deletes had zero tests

Thirteen tests pinned `find_`, which touches nothing. `purge_` — the shipped function, the one containing `shutil.rmtree` — had none. A reviewer proved it with two mutants that shipped green:

| mutant | before | now |
|---|---|---|
| `purge_` iterates the whole parent dir instead of `find_`'s output — one refactor from `rm -rf` on every app in `/Applications` | 24 passed | **3 red** |
| `purge_` deletes nothing at all | 24 passed | **3 red** |
| the skip-on-failure `continue` removed | — | **2 red** |
| version segment broadened `[0-9][0-9.]*` → `.*` | 24 passed | **3 red** |

### A rationale that was simply false

The docstring said the sweep runs after `_maybe_relocate_to_applications` "because that call can move us". It doesn't: if it moves us it ends in `os._exit(0)`, so the sweep never runs in that process. The ordering actually means the sweep only runs when we were **not** relocated — so it can be operating in `~/Downloads`, a mounted DMG, or a Gatekeeper `AppTranslocation` path. Acceptable (every candidate must still be a sibling, match the closed list, and claim our bundle id) but not what the comment claimed, and that is the kind of sentence a later reviewer reasons from.

### Also fixed

- **Case sensitivity.** `/Applications` is APFS and case-**in**sensitive, so `betterflow.app.old` and `BetterFlow.app.old` are one file to the filesystem and were two strings to the matcher. Widened with `IGNORECASE`; the running bundle still cannot self-match under any casing.
- **`.old.app` was spelled twice** in one module, 640 lines apart — the writer and an independent regex in the sweep. Now one constant. The other two shapes stay unlinked deliberately: the AppImage form cannot produce a `.app.old`, and `-<ver>-backup` has no producer in this tree, so linking them would assert a provenance the code cannot support.
- **The bundle-id guard missed the authority.** What the sweep compares against is `CFBundleIdentifier` from `build.spec`, not any Python constant. Change `build.spec` alone and the old guard stayed green while the sweep matched nothing forever *and* the TCC insert wrote a grant no installed app claims — the other #211 symptom. The launchd-label assertion is split out and documented as a **coincidence**, not an invariant, since Apple's convention allows `<bundleid>.agent` and a legitimate change there must not redden a test labelled as protecting the sweep.
- **Three derivations of "the running bundle path"**, only one of which resolves symlinks — measured diverging through a symlinked ancestor, different directory scanned, silently. The sweep now reuses the module's own resolver, via a public alias rather than a rename (dropping the underscore broke 10 call sites in three unrelated test files).
- **The callsite grep** covered all 4,200 lines of `main.py` and had three demonstrated defeats. Replaced with behavioural tests of the extracted method plus a grep scoped to `run()`'s own source.

### One mutant still survives, and one did until just now

`entry == running_app` survives and is kept — unreachable while the name patterns are correct, proven load-bearing as a pair, documented at both sites.

The non-canonical **warning** also survived the matrix, which mattered: its return value is identical with and without the bail-out, so only the log line distinguishes them, and nothing asserted it. That is the same shape as #218's wake/unlock re-asserts — the fix for a reviewer's headline finding shipping with nothing watching it. Now witnessed, with a control that the canonical copy stays quiet.

### Worth recording

A reviewer built two real `.app` bundles sharing a bundle id, registered both with `lsregister`, and **could not reproduce** `open -a <path>` launching anything other than the path given, on macOS 15.x. The comment in `autostart.py` already labels that mechanism INFERRED. That labelling is correct and stays — the change removes a degree of freedom for free, and nobody should later read it as the proven cause.

**Suite: 1938 passed, 1 skipped, exit 0.** ruff at baseline.

tudorsss:
## Round 3 — refutation on the cumulative diff

The shipped code held under ~20 mutants. **Three guards did not**, and one of them was a test written in round 2 specifically to close a gap — it passed for the wrong reason.

### The darwin gate had no witness

`test_the_sweep_does_nothing_off_macos` patched `sys.platform` to `win32` but not the resolver, unlike its three siblings. So the real resolver took its own `win32` branch, found `sys.frozen` false, returned `None`, and the `running is None` guard produced the identical `called == []`. **Removing the platform check entirely left the whole suite green.**

That matters beyond tidiness — measured with the gate gone:

```
linux  /opt/BetterFlow-linux-x86_64.AppImage
  -> WARNING "Running from 'BetterFlow-linux-x86_64.AppImage', not the
     canonical BetterFlow.app ... (#211)"
```

Every Linux AppImage start would emit the non-canonical warning — the exact fleet-discovery signal this branch exists to create, turned into pure noise.

### The branch made `build.spec`'s bundle NAME load-bearing and left it unpinned

`_CANONICAL_STEM` is now the switch deciding whether the sweep runs at all, and its authority is `build.spec`'s `name=`/`CFBundleName`. Renaming both there left the full suite green while silently disabling the sweep on every device forever *and* making every device log the alarming warning. This repo has already done exactly one such rename — which this branch's own legacy-name test cites.

The guard that pins the bundle *id* reasons about precisely this class in its docstring, and I did not apply it to the second authority the same commit introduced.

### The shared-constant test proved nothing about the writer

Round 2 added a test to prove the writer and the sweep agree about `.old.app`. It rebuilt the expected name **from the constant**, so both sides of the assertion traced back to one artifact, and reverting the writer to a literal `.bak.app` changed nothing it could see. Rule 6a, inside the test written to close that gap.

`_macos_backup_path()` now owns the expression and `_apply_macos_update` calls it, so the test asks the writer instead of restating it. Scope stated in the docstring: it exercises the helper, not `_apply_macos_update`, which needs a downloaded archive to run.

### Also closed

- `test_..._logs_what_went` asserted only the call, not the log — deleting the whole `logger.info` block survived. Phantom 9: a verb in the name with nothing behind it.
- `get_app_bundle_path` was `= _get_app_bundle_path`, which binds the function object at import time, so a test patching the private name silently got the real resolver. Three existing files patch only that name. Now a wrapper, **and** witnessed — reverting it survived every other test here, because they all patch both names.
- A docstring claiming the case-insensitivity fixture "only reaches the branch on a case-sensitive filesystem" was simply wrong; it runs and genuinely witnesses `IGNORECASE` here.
- The legacy-name test asserted the forward direction while its docstring described the reverse. The reverse now has its own test.

### Matrix, re-run against a committed tree

```
darwin gate removed              -> test_the_sweep_does_nothing_off_macos           RED
build.spec bundle name renamed   -> test_the_identity_..._once_per_reason           RED
backup writer -> literal         -> test_the_name_the_updater_leaves_behind...      RED
removal log deleted              -> test_..._passes_the_resolved_bundle_and_logs... RED
public wrapper -> assignment     -> test_the_public_resolver_follows_a_patch...     RED
```

Each mutant reddens a **distinct** named test; control clean before and after; tree restored to zero dirty files.

### Negative results worth keeping

The reviewer checked **all 16,384 casings** of `BetterFlow.app` against the three patterns — zero match, so the `IGNORECASE` self-match claim in the comment is proven rather than argued. And a partial `rmtree` leaves `Contents/Info.plist` intact, so an interrupted sweep stays identifiable and the next one re-finds it: retry-safe.

**Suite: 1941 passed, 1 skipped, exit 0.** ruff at baseline.

PR #224, opened 2026-08-25
Cleanup failures should not abort an update, or bury the real cause

tudorsss:
The actual trigger of #211, and the half [#219](https://github.com/Better-Quality-Assurance/betterflow-sync/pull/219) deliberately did not take.

## What happened

Martin's Mac, 2026-08-23. An update from 1.5.119 to 1.5.125 died on:

```
hdiutil detach failed (rc=1)
Update failed: Failed to detach DMG mount - aborting update
```

The machine then ran **1.5.119 for days** — below the server's own minimum version floor — logging `update required` on every sync.

## The `.app` had already been copied out

That is the whole finding. `shutil.copytree` completes inside the `try`; the `RuntimeError` was raised from the `finally`, *after* the work succeeded. Unmounting a disk image is housekeeping, and the trade is not close:

| | cost |
|---|---|
| failed unmount | a stale entry in `/Volumes` until the next reboot |
| the abort | a fleet device stuck on an old build |

Worse, the abort stranded `staged_update/`, so the retry loop then failed on a missing `staged.json` and reported **that** instead — burying the real cause one layer further. That is the second error in the incident log, and it is a consequence, not a separate fault.

## A second defect, invisible in the logs

`raise` inside `finally` **replaces the exception in flight**. A genuine failure in the copy above would have been relabelled as a detach problem and its cause discarded. A naive fix of the first defect leaves this one standing, so it gets its own test: with no `.app` in the image, the caller must still see `FileNotFoundError`, not a `RuntimeError` about unmounting.

## Proof of failure

Captured before any edit:

```
test_a_failed_detach_keeps_the_extracted_app             FAILED
test_it_retries_the_detach_with_force                    FAILED
test_a_detach_that_never_succeeds_still_does_not_abort   FAILED
test_a_timeout_on_detach_is_also_not_fatal               FAILED
test_a_real_failure_is_not_relabelled                    FAILED
test_the_happy_path_still_detaches_exactly_once          passed   <- control
```

The control passing pre-fix is the point — the tests isolate the defect rather than describe the new code.

## The fix

`_detach_dmg` makes two attempts and **never raises**. The first is polite; the retry adds `-force`, which clears the usual cause — a straggling process still holding the volume. Asserted on the **arguments**, so a retry that merely repeats the same failing command does not pass. The happy path still detaches exactly once, because making failure non-fatal must not mean leaking a volume on every update.

## Mutation matrix

Five mechanisms, each reddening distinct named tests. Control clean before and after; tree restored to zero dirty files.

| mutant | witnesses |
|---|---|
| restore the `raise` (the original defect) | `keeps_the_extracted_app`, `never_succeeds_still_does_not_abort`, `real_failure_is_not_relabelled`, `timeout_is_also_not_fatal` |
| drop the `-force` escalation | `it_retries_the_detach_with_force` |
| no retry at all | `it_retries_the_detach_with_force` |
| skip detach entirely | `happy_path_still_detaches_exactly_once`, `it_retries_the_detach_with_force` |
| timeout fatal again | `a_timeout_on_detach_is_also_not_fatal` |

## Verification

- `pytest tests/` → **1952 passed, 1 skipped, exit 0**. That is the whole of CI's `test` job.
- `ruff` → *All checks passed* on both touched files.
- Not smoke-tested against a real DMG that refuses to unmount — reproducing that needs a process holding the volume on a real Mac. The `hdiutil` calls are driven through a stub that materialises a real mount directory, so the copy under test operates on real files; the `-force` escalation itself is verified by argument, not by observing macOS accept it.

## Still open on #211

Alerting ops when a device sits below the server's minimum version floor for more than a day. The signal reached the server throughout this incident and nobody was paged — that is the part that would have caught it in hours rather than days.

Refs #211.

### DIFF
--- a/src/self_updater.py
+++ b/src/self_updater.py
@@ -600,9 +600,11 @@ def _find_app_in(directory: Path) -> Optional[Path]:
     return None
 
 
-# Polite first, then force. `-force` is what clears the usual cause of a failed
-# unmount: a straggling process still holding the volume.
-# Polite first, then force.
+# Polite first, then force. See _detach_dmg for what `-force` does and does not
+# buy — it fixes rc=16 (a holder) and does NOT help rc=1 (never mounted), which
+# is the code #211 actually logged. This comment used to assert the rc=16 story
+# as that incident's cause; it was refuted by measurement and the correction
+# landed 45 lines below without touching the text a reader meets FIRST.
 #
 # No `-quiet`: output is captured, not printed, so it bought nothing and it
 # BLANKED the only diagnostic we get. Measured on Darwin 24.6.0 — with `-quiet`
@@ -709,12 +711,36 @@ def _extract_from_dmg(dmg_path: Path, extract_dir: Path) -> None:
         # Mount DMG read-only, hidden from Finder.
         # NOTE: no -noverify — we want hdiutil to verify the DMG checksum at
         # mount time. Signature verification still happens post-extract.
-        subprocess.run(
-            ["hdiutil", "attach", "-nobrowse", "-readonly",
-             "-mountpoint", str(mount_point), str(dmg_path)],
-            capture_output=True, text=True, timeout=60, check=True,
-        )
+        #
+        # The attach gets its OWN try so the two failure modes can be told apart;
+        # they need opposite cleanup.
+        try:
+            subprocess.run(
+                ["hdiutil", "attach", "-nobrowse", "-readonly",
+                 "-mountpoint", str(mount_point), str(dmg_path)],
+                capture_output=True, text=True, timeout=60, check=True,
+            )
+        except subprocess.TimeoutExpired:
+            # hdiutil was SIGKILLed by subprocess.run's own timeout, mid-flight.
+            # It may have attached first, and SIGKILL runs no cleanup handler —
+            # so the image can be live with nothing left to release it.
+            #
+            # Witnessed on a real 400 MB DMG by sweeping the timeout: at
+            # t=0.05/0.10/0.15s the process died with the image STILL ATTACHED,
+            # 3 leaks in 8 runs. rmtree cannot recover it (a live mount point
+            # survives rmtree), so mount and temp dir persist until reboot.
+            #
+            # origin/main detached here unconditionally. Gating on `mounted`
+            # alone made this a REGRESSION against it — the gate was right about
+            # the cause and wrong about the discriminator.
+            mounted = True
+            raise
+        # A CalledProcessError needs no clause: hdiutil DECLINED cleanly (a
+        # corrupt or truncated download failing its checksum verify), nothing
+        # attached, so `mounted` stays False and the detach is skipped. That is
+        # the case the gate exists for.
         mounted = True
+
         # Find .app inside mounted volume
         app_found = None
         for item in mount_point.iterdir():
--- a/src/self_updater.py
+++ b/src/self_updater.py
@@ -637,8 +637,25 @@ def _detach_dmg(mount_point: Path) -> bool:
     copy above would have been relabelled as a detach problem, and its cause
     discarded.
 
-    Two attempts. The first is polite; the retry adds `-force`, which is what
-    clears the usual cause — a straggling process still holding the volume.
+    Two attempts. The first is polite; the retry adds `-force`.
+
+    Be precise about what `-force` does and does not buy, because an earlier
+    version of this docstring got it wrong. Measured on Darwin 24.6.0:
+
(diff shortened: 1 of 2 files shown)


tudorsss:
## Correction: my stated mechanism was wrong, and I could not verify the headline claim

Two reviewers went over this. The design lens caught two invented facts; the correctness lens caught something bigger.

### `-force` would not have fixed #211

A reviewer measured `hdiutil` on Darwin 24.6.0:

| condition | rc | `-force` helps? |
|---|---|---|
| path is not a mount / never was | **1** | **no — still rc=1** |
| volume busy (a holder) | 16 | yes (16 → 0) |

**The incident logged rc=1.** So the "straggling process still holding the volume, cleared by `-force`" story — which I put in the docstring, the commit message and this PR body — is not that incident's mechanism. rc=1 fits `attach` having failed first, a corrupt or truncated download failing hdiutil's checksum verify.

That makes the **`mounted` gate** the load-bearing change here, not the retry.

### I tried to settle it and could not

Requested a log capture from device 14. `betterflow.log` had already rotated past 2026-08-23, so the one line that would decide it — `Extracted BetterFlow.app from DMG`, immediately before the failure — is gone. `self-update-relaunch.log` survives back to June and shows the last relaunch on **2026-08-21**, nothing on 08-23: consistent with an abort, but it cannot distinguish *which* abort.

So the original headline, *"the .app had already been copied out"*, is now marked **inferred** rather than asserted. It is true of the code path in general and unproven for this incident. The fix stands under either reading — if the copy succeeded it saves the update; if attach failed it stops us burying the real cause under a false unmount error.

### Two more real defects, both mine

**`_detach_dmg` promised "never raises" and did.** It caught only `TimeoutExpired`. `subprocess.run` also raises the `OSError` family *before the child starts* — `FileNotFoundError` when hdiutil is unresolvable, `ENOMEM`/`EAGAIN` when `posix_spawn` fails under pressure in a long-lived multi-threaded tray app, `PermissionError` under a sandbox. From a `finally` that restored **both** of #211's defects. The reviewer's three probes are kept as permanent tests.

**`-quiet` blanked the only diagnostic.** Measured: with it, both rc=1 and rc=16 return empty stderr, so `"rc=%d: %s"` always logged an empty `%s`. Output is captured rather than printed, so the flag bought nothing and cost the diagnostic — in a fix whose whole premise is that #211 cost days partly to diagnostic poverty.

**Detach ran even when attach failed** (design lens), logging a confident "could not unmount" *above* the real cause. Now gated on `mounted`; the `rmtree` stays unconditional, since gating it too would leak an empty temp dir on every attach failure.

### Two facts I invented

- *"a stale entry in `/Volumes`"* — wrong. With `-mountpoint <mkdtemp>` and `-nobrowse` the leak lives under `/var/folders` and is invisible in Finder. An operator told to check `/Volumes` finds nothing and concludes it did not happen.
- *"the abort strands `staged_update/`, so the retry fails on a missing `staged.json`"* — **false**, and it contradicts this file's own module docstring. `apply_staged_update` calls `clear_staged_update()` **before** `_apply_local_artifact`, deliberately, and `get_staged_update` answers a missing meta file with a silent `return None`. The incident log showed those two errors in sequence and I inferred a chain between them. The true cost is a better argument anyway: the `finally` rmtree's the artifact, so the abort discards the whole download.

### Matrix, re-run against a committed tree

```
catch TimeoutExpired only  -> 6 red, incl. all three OSError probes
restore -quiet             -> test_the_detach_command_keeps_its_stderr
detach ungated             -> test_a_failed_attach_does_not_try_to_unmount_anything
```

Earlier round: restoring the `raise` reddens 4; dropping `-force` and skipping detach each redden their own. Controls clean either side, tree restored.

**Suite: 1963 passed, 1 skipped, exit 0.** ruff clean.

### Known and deliberately not fixed here

- **The leaked mount pins unreclaimable disk** — the reviewer measured a real DMG at ~60 MB held open after `unlink`, worst case ~2.9 GB/day across 30-minute retries until reboot. Nothing sweeps `betterflow-dmg-mount-*`. That is a fourth mechanism for #211 and belongs in its own change.
- **`-> bool` is still read by nobody.** Now asserted in tests so it is not a free mutation, but the useful consumer would be a `leaked_mount` counter on the heartbeat — which is the same ops-visibility work #211's remaining bullet describes.
- **`_detach_dmg` / `rmtree` ordering is unwitnessed.** Harmless while the attach is `-readonly` (verified: rmtree on a live read-only mount deletes nothing); a latent hazard if `-readonly` ever leaves.

Issue #223, opened 2026-08-25
A backed-off download reports the device as started without clearing the capture-dead flag


BetterQAFTW:
## Business impact

A Mac whose tracker download has failed keeps reporting itself as "not recording" to the fleet even after something else starts capturing on its port. Support sees a device flagged as capture-dead that is in fact recording fine, and the longer that persists the more the ops ingest learns to ignore the flag — which is the signal that exists to catch genuinely silent machines.

The window is not small. The download retry interval escalates to an hour, so a stuck device spends nearly every 60-second cycle on this path rather than on a real download attempt.

## Steps to reproduce

1. Put a device in the state where the tracker download fails and no server is on port 5600. `tracker_download_failed` latches.
2. Start something on port 5600 (a hand-installed ActivityWatch, for example).
3. Let the next cycle run while the download backoff is still in effect.

## Expected result

The device is recording, so it stops reporting `tracker_download_failed`, and `_using_external` reflects that it attached to a server it did not start.

## Actual result

`_start_locked` returns `True` — so the caller believes the tracker started — while `tracker_download_failed` stays latched and `_using_external` stays `False`. The heartbeat keeps reporting capture-dead.

## Technical details

Verified against `origin/main` on 2026-08-25, `src/aw_manager.py:1128-1138`:

```python
if since_last < self._download_retry_interval:
    # Backed off. Report the same outcome the last real attempt did:
    # an attached external server still captures, nothing else does.
    logger.debug(...)
    self._managed_components_unavailable = True
    return server_already_running          # <- no clear, no _using_external
```

The comment states the intent correctly — *"report the same outcome the last real attempt did"* — and the code does not carry it out. The three sibling attach sites all clear the flag (`:1086`, `:1121`, `:1144`); this one returns the bare boolean.

Two notes for whoever picks it up. `server_already_running` comes from `_port_in_use()`, a bare TCP connect, whereas the sibling sites were deliberately moved onto `_server_responding()` (a real `/api/0/info` ask) precisely because a dead `bf-data-service` holding the socket is not a capturing device — so the fix should adopt that predicate rather than trust the port check. And this sits inside `_start_locked`, which #217 rewrites substantially (297/82 in that file), so it is better done after that lands than in a race with it.

Found by the pre-merge gate on #213; the other two findings from that gate shipped in #222.

## Environment

[...]

PR #230, opened 2026-08-26
A backed-off download must not report a recording Mac as capture-dead (#223)

BetterQAFTW:
## What this fixes

The backoff branch in `_start_locked` says in its own comment:

> Report the same outcome the last real attempt did: an attached external
> server still captures, nothing else does.

It did not carry that out. It returned the bare `server_already_running`
boolean and left `tracker_download_failed` latched and `_using_external`
False, so a Mac attached to a healthy external tracker kept reporting itself
capture-dead to the fleet. The retry interval escalates to an hour, so a
stuck device spends nearly every 60-second cycle on this path rather than on
a real download attempt.

## The part that is easy to get wrong

The clear is gated on `_server_responding()` (a real `/api/0/info` ask), NOT
on `server_already_running` (a bare TCP connect) — matching the Rosetta
attach path forty lines above, which pays for the same HTTP ask for the same
reason.

A held socket proves a process, never a capture. Clearing the flag on a
held-but-dead port would report a **silent** device as healthy, which is
strictly worse than the bug being fixed, because that is precisely the
device this flag exists to catch.

## Verification

| check | result |
|---|---|
| new tests pre-fix | **3 failed, 2 passed** |
| new tests post-fix | 5 passed |
| mutant: gate on the bare TCP connect instead of the HTTP ask | **kills 3 named tests**, including the held-but-dead fail-open guard |
| full suite | 2023 passed, 1 skipped, 0 failed |

The pre-fix split is the useful part: only the three defect-specific cases
fail, while the two that pin already-correct behaviour (dead port, empty
port) pass both before and after. That is a proof of the bug rather than a
spec for the new code.

## Not covered

When the port is held but dead, this branch still returns `True`, so the
caller believes a tracker started. That contradicts the same comment's
"nothing else does" and is arguably a second defect — but changing the
return value touches every caller of `_start_locked`, so it is deliberately
left alone here and worth its own issue.

### DIFF
--- a/src/aw_manager.py
+++ b/src/aw_manager.py
@@ -1287,6 +1287,23 @@ class AWManager:
                     self._download_retry_interval,
                 )
                 self._managed_components_unavailable = True
+                if server_already_running and self._server_responding():
+                    # The comment above says "an attached external server still
+                    # captures" and this branch did not carry it out: it left
+                    # tracker_download_failed latched and _using_external False,
+                    # so a Mac that IS recording kept reporting itself
+                    # capture-dead to the fleet -- for nearly every 60s cycle,
+                    # because the retry interval escalates to an hour (#223).
+                    #
+                    # Guarded on _server_responding(), NOT on the bare
+                    # server_already_running TCP connect, for the same reason
+                    # the Rosetta attach path forty lines up is: a held socket
+                    # proves a process, never a capture. Clearing the flag on a
+                    # held-but-dead port would report a silent device healthy,
+                    # which is the failure this flag exists to catch.
+                    self._using_external = True
+                    self.tracker_download_failed = False
+                    return True
                 return server_already_running
             self._last_download_attempt = now
             logger.info("Tracker components not found, downloading...")
--- a/tests/test_aw_manager_backoff_attach.py
+++ b/tests/test_aw_manager_backoff_attach.py
@@ -1,0 +1,98 @@
+"""A backed-off download must not report a recording Mac as capture-dead (#223).
+
+When the tracker download has failed and the retry is in backoff, the branch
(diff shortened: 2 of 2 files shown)


BetterQAFTW:
Verified independently before merging, on Python 3.12 (CI pins 3.11; 3.11 is not on this machine, 3.12 is the closest available and the local default 3.14 is further from CI than either).

**Proof of failure reproduced.** `src/aw_manager.py` restored to `origin/main` via `git show`, `__pycache__` cleared first so a same-size replacement could not be served from stale bytecode:

```
FAILED  TestBackoffWithACapturingServer::test_clears_the_capture_dead_flag
FAILED  TestBackoffWithACapturingServer::test_records_that_it_attached_to_a_server_it_did_not_start
FAILED  TestBackoffWithoutACapturingServer::test_the_clear_is_gated_on_the_http_ask_not_the_tcp_connect
3 failed, 2 passed
```

Matches the PR's claim exactly, and it is the right shape: only the three defect-specific tests fail, while the two describing already-correct behaviour pass. The third failure is the fail-open guard — `the branch must pay for the HTTP ask before granting an attach` — which is the assertion that matters most here, since clearing on a held-but-dead port would report a silent device as healthy.

Because main had moved 3 commits ahead I ran the suite on the **merged** tree rather than the branch:

| check | result |
|---|---|
| `pytest tests/` on the branch | 2023 passed, 1 skipped, 0 failed |
| `pytest tests/` on the merged tree | 2034 passed, 1 skipped, 0 failed |
| ruff, two touched files | 4 findings — the same 4 `origin/main`'s `aw_manager.py` already has, so 0 added (and ruff is in no workflow here) |

Merging.

PR #234, opened 2026-08-31
Record the server's rejection reason on dead-lettered events, and stop the suite notifying the machine that runs it

tudorsss:
Two changes from one overnight sweep finding. Separate commits; the second is unrelated to the first except that it was found while verifying it.

## 1. The server's rejection reason had nowhere to go

An error sweep surfaced repeated warnings from the fleet:

```
Dropped 1 queued event(s) after max retries — the server rejected them;
held in dead-letter for replay, not discarded (buckets=bf-status, oldest ~3659m old)
```

**Not data loss.** The events are preserved, and the drop→requeue→drop churn is by design: `requeue_storable_dead_letter` resurrects storable rows every cycle with a fresh retry budget, so a genuinely-poison event re-drops every ~35 min (1800s cooldown + 5 retries) until it ages past `EVENT_RETENTION_DAYS = 7`. The docstring says so.

**What could not be answered was WHY the server refused** — malformed event, or a receiving-side change? Those have opposite fixes. Every dead-lettered row carried the agent's own generic string, `"exceeded max retries (5); definitive rejection"`, which describes what *we* did. The server side was silent too: zero `error_logs` entries for batch rejections in 7 days, because a validation 422 raises no Laravel exception.

`send_events` already captured the server's text on `SyncResult.error`. **Nothing downstream ever read it.** The break was one link further down than the obvious reading suggests.

The reason must be recorded when the batch FAILS and read when the event is DROPPED, and those are different cycles — `remove_failed` runs at the top of `_process_queue`, decoupled from the batch that failed — so it needs somewhere to live in between:

- `queued_events` gains a nullable `last_error`, PRAGMA-guarded additive migration. Unread by delivery, so a downgrade is safe.
- `increment_retry(ids, last_error)` records it, **only when supplied**, so a later reason-less failure cannot erase a captured 4xx.
- `remove_failed` prefers the row's reason over the generic string.
- The transient stuck-head shed passes an explicit description instead, so a timeout can never masquerade as a server rejection.

### The ops report gets the status code ONLY

That warning goes to the **cross-tenant** ingest, and a server validation error routinely echoes the value it rejected — our payloads carry window titles. Full text stays in the local dead-letter row where an engineer pulls it deliberately. Same reasoning the function already uses to reduce bucket ids to types (privacy F4), and an **allowlist** rather than a blocklist because the ways free text can carry a title are unbounded (`fail-closed.md` Rule 5). Three tests assert a title-bearing error never reaches the message.

## 2. The suite was posting real notifications

Found by running the suite five times while verifying the above: it buried the developer's Notification Center in ~50 banners. `send_notification` falls back to `osascript` when pyobjc is absent (a normal dev venv), so they arrive attributed to **Script Editor** — which `clear_notifications()` documents it *cannot* clear programmatically. **10 per full-suite run**, now reported by a terminal-summary line so the leak stays visible if it returns.

Same class as the config incident this `conftest` already guards ("This hit a real machine on 2026-07-14. It is not a hypothetical").

Patches the **private** senders, not `send_notification()` — callers bind that name at import time (`src/main.py` says so explicitly), so patching the public function misses every module-level importer.

## Verification

**Proof of failure (change 1):** `tests/test_dead_letter_reject_reason.py` written first. Pre-fix 2 of 3 failed on `TypeError: increment_retry() got an unexpected keyword argument 'last_error'`, while `test_falls_back_to_generic_when_server_gave_no_reason` already passed and pins existing behaviour. Post-fix all 6 pass.

**Guard witnessed failing (change 2):** with the fixture short-circuited, both guard tests fail, and the mutant run's own log carries the real sender being reached (`notifications.py:222`) — independent evidence the egress path is live. Both pass once restored.

**Full suite: 2037 passed, 1 skipped, exit 0**, and silent.

One flake seen and ruled out: `test_the_re_probe_does_not_reach_the_sixty_second_start_path` failed once under load average ~50, passed 19/19 in isolation and in three subsequent full runs. Not related to this branch.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,9 @@ ignore = ["E501"]
 testpaths = ["tests"]
 python_files = ["test_*.py"]
 addopts = "-v --tb=short"
+markers = [
+    "real_notifications: test exercises a notification sender's own internals; exempt from the conftest block (it mocks the OS boundary itself)",
+]
 # `from src.* import ...` works today only because pytest's default
 # rootdir-on-sys.path heuristic happens to add the repo root. Pin it
 # explicitly so CI runners or invocations from inside tests/ also
--- a/src/main.py
+++ b/src/main.py
@@ -37,6 +37,7 @@ try:
     )
     from .reminders import ReminderManager
     from .sync import AWClient, BetterFlowClient, OfflineQueue, SyncEngine
+    from .sync.sync_engine import server_status_summary
     from .sync.http_client import BetterFlowAuthError, transient_failure_counter
     from .sync.os_idle import get_system_idle_seconds
     from .system_events import start_system_event_listener
@@ -72,6 +73,7 @@ except ImportError:
     )
     from reminders import ReminderManager
     from sync import AWClient, BetterFlowClient, OfflineQueue, SyncEngine
+    from sync.sync_engine import server_status_summary
     from sync.http_client import BetterFlowAuthError, transient_failure_counter
     from sync.os_idle import get_system_idle_seconds
     from system_events import start_system_event_listener
@@ -2161,12 +2163,23 @@ class SyncCoordinator:
         re-login is expected, not a failure worth alerting on.
         """
         self._consecutive_sync_failures += 1
+        # Full text to the LOCAL log only. `reason` is the server's own response
+        # body, passed through verbatim by http_client, and a validation error
+        # routinely echoes the value it rejected — which for us can be a window
+        # title. betterflow.log stays on the device and is uploaded only on an
+        # explicit admin request; the ops ingest below is cross-tenant and
+        # always on, so it gets the status code and nothing else.
+        logger.warning(
+            "Sync failing repeatedly (%d×): %s",
+            self._consecutive_sync_failures, reason,
+        )
         if (
             self.error_reporter is not None
             and self._consecutive_sync_failures >= self._SYNC_FAILURE_ALERT_THRESHOLD
         ):
+            safe = server_status_summary(reason).lstrip("; ") or "no server status"
             self.error_reporter.capture(
-                f"Sync failing repeatedly ({self._consecutive_sync_failures}×): {reason}",
+                f"Sync failing repeatedly ({self._consecutive_sync_failures}×): {safe}",
                 level="error",
                 exc=exc,
                 tags={"component": "sync"},
--- a/src/sync/bf_client.py
+++ b/src/sync/bf_client.py
@@ -605,7 +605,8 @@ class BetterFlowClient(BaseApiClient):
 
         Privacy note: this is an admin-initiated diagnostic pull. The log can
         contain app names, the device hostname (in bucket ids), and OS usernames
-        in stack-trace paths — but NOT window titles, URLs, or auth tokens
+        in stack-trace paths, and server RESPONSE text (which can echo a value the
+        server rejected, including a window title) — but NOT auth tokens
         (those are never logged). Acceptable within the tenant-admin trust model.
         """
         files: dict = {"log": ("betterflow.log", log_tail, "text/plain")}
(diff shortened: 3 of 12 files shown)


tudorsss:
## Review round complete — 8 findings fixed, C1 resolved by decision

Two parallel reviewers (correctness + design/privacy). Both independently found the same two defects, which is what raised confidence in them.

### Fixed

| Finding | What was wrong |
|---|---|
| **Provenance, not shape** | The sanitizer searched free text for any 3-digit run. `"shed as unstorable after 137 transient failures — not a server rejection"` → `server status 137`; `"Bug 404 fix.txt"` → `server status 404`. Now anchored on `^API error \((\d{3})\)`, the only shape http_client produces for a definitive rejection. |
| **Multi-cause drops** | `failed_event_summary` picked one arbitrary row's reason from a table-wide predicate. My comment claiming those rows "share a batch" was false — they accumulate across batches and cycles. Now returns the distinct set; the message names all of them. |
| **Per-event path recorded nothing** | `result.error` is `None` on the partial-accept branch, so the one path where the server names a *specific* event wrote no reason — and never-overwrite-with-`None` meant it kept a stale string from an unrelated earlier whole-batch failure. |
| **Eviction erased reasons** | `evict_unstorable` still selected 4 columns and wrote its generic string over a real 4xx. Now shares `remove_failed`'s precedence. |
| **Unbounded text** | Server-controlled `message` written verbatim to every event in a failing batch; truncated at the write. |

### C1 — the leak this PR did not touch

Raw `result.error` already reached the same cross-tenant ops ingest by another door: `sync_engine:3486` → `main:1896` → `main:2168`, verbatim, on the 3rd consecutive failure. Sanitizing the drop report while documenting the invariant as held was worse than not claiming it.

**Resolved by decision:** full text stays in `betterflow.log` (on-device, uploaded only on explicit admin request); the always-on ops ingest gets the status code. An outage is still fully diagnosable, at the cost of one request. The sanitizer is now one module-level `server_status_summary()` shared by both call sites rather than two spellings.

`upload_logs`' docstring claimed the log holds "NOT window titles". That was **already false** (`sync_engine:3478`) and this change relies on it deliberately, so it is corrected rather than left to mislead whoever next decides a log upload is safe.

### Test gaps both reviewers found — the more useful half

- **The sanitizer call was unwitnessed.** All three privacy tests called `_dropped_reason_code` directly, so deleting the call site — or replacing it with the raw string, i.e. the exact leak — kept the suite green. Now driven through `_report_dropped_events` with a fake reporter.
- **The migration had zero coverage.** Deleting both `ALTER` lines left every test green, because a fresh `OfflineQueue` already `CREATE`s the column. Now built on a real 4-column legacy DB with a pre-existing row.

### Mutation matrix — 8 mechanisms, 8 distinct named killers, clean control

```
sanitizer call site deleted        -> TestSanitizerIsActuallyCalled
sanitizer call leaks raw text      -> TestSanitizerIsActuallyCalled
regex unanchored (free-text)       -> TestReasonProvenanceNotShape (x2)
migration ALTER deleted            -> TestLegacyQueueMigration
evict precedence reverted          -> TestReasonSurvivesTheOtherTwoDropPaths
remove_failed precedence reverted  -> TestDeadLetterCarriesServerReason
distinct set collapsed             -> TestSummaryCollectsEveryDistinctCause
sync-failure alert un-sanitized    -> TestSyncFailureAlertDoesNotEchoServerText
```

One mutant survived the first pass: collapsing the distinct set reddened nothing, because the multi-cause test drove the *renderer* and not the *collector*. Fixed and re-witnessed — the same producer/consumer split the reviewers had just flagged one level up.

**Full suite: 2047 passed, 1 skipped, exit 0.**

### Verified clean by review, unchanged

`remove_failed` row indices; no `SELECT *` on `queued_events`; downgrade safety; `evict_unstorable` KeyError (guarded by `.get`); cross-platform conftest behaviour; and the `event_data`-vs-column design choice (storing the reason in `event_data` would put it on the wire via the replay path).

PR #235, opened 2026-08-31
chore(release): 1.5.129

tudorsss:
Version bump only — one line in `src/__init__.py`.

Ships **#234** to the fleet:

- The server's 4xx rejection reason now survives to the dead-letter row. It was captured on `SyncResult.error` and read by nothing, so every dead-lettered event recorded the agent's own generic string and "malformed event or receiving-side change?" was unanswerable from either end.
- The ops warning carries the HTTP status only (`server_status_summary`, anchored on `^API error \((\d{3})\)`); full server text stays in `betterflow.log`, which is upload-on-admin-request.
- The test suite no longer posts real notifications to the machine running it — 10 per full-suite run via the osascript fallback, attributed to Script Editor and unclearable programmatically.

Until this tag lands, all of that is on `main` and on zero laptops: releases here are tag-driven, so the merge reached nobody.

## Pre-release checks

- Clean checkout of the release tree: **2052 passed, 1 skipped, exit 0** (2053 collected).
- Tree verified identical to the merge commit `6aa3e25` apart from the version line.
- **Cross-platform risk specifically checked**, because this repo has broken three releases on tests that pass locally and fail on another platform, and the new conftest fixture patches four platform-specific senders with `raising=True`. Parsed `src/notifications.py` with `ast` (not grep): all six senders — `_send_macos_pyobjc`, `_send_macos_osascript`, `_send_windows`, `_send_linux`, `_clear_macos_pyobjc`, `_clear_windows` — are unconditional module-level defs, so `monkeypatch.setattr(..., raising=True)` cannot fail on any platform. The ubuntu PR gate only exercises one of the three the tag build runs.

## Note on a number

An earlier full-suite run on the pre-merge branch reported 2047 passed. The clean checkout reports 2052. My files hold exactly the 18 tests expected and no commits landed on `main` between the merge and this branch, so the released tree is not in question — but I could not reproduce the lower figure, since it came from a worktree that had been through many mutation cycles and has since been removed. Recording it rather than inventing a cause.


tudorsss:
## Correction: the test-count gap is explained, and this release ships more than #234

The note above said the 2047 → 2052 difference could not be reproduced and declined to guess a cause. It now has a verified one, so the note is wrong rather than merely incomplete.

**#230 ("A backed-off download must not report a recording Mac as capture-dead", #223) merged at 09:09:52Z — an hour before #234 merged at 10:16:48Z.** My branch was cut from `b45014c` before that landed, so it never carried #230's tests; `main` does.

It adds `tests/test_aw_manager_backoff_attach.py`, and the net new test-function count is **exactly 5** — precisely the gap:

```
2047 (my pre-merge branch)  +  5 (#230)  =  2052 (clean checkout of main)
```

Nothing about the released tree changes, and nothing was wrong with either measurement — they were taken against different bases, which is the one possibility I did not check before writing "could not reproduce".

**The consequence that matters: this release is not #234 alone.** `v1.5.129` will ship both:

- **#234** — the server's rejection reason survives to the dead-letter row; ops warning carries the HTTP status only; the test suite no longer posts real notifications.
- **#230** — a backed-off tracker download no longer reports a recording Mac as capture-dead.

I did not review #230 and am not vouching for it; flagging it because the PR body reads as though this tag ships one change, and whoever reads it during a future bisect deserves to know it carries two.

PR #236, opened 2026-08-31
Terminal dead letters permanently starved the replay (silent loss of billable activity)

tudorsss:
**This is silent loss of captured activity, not a metrics problem.** Found while a read-only agent was analysing something else entirely; reproduced independently before acting on it.

## The defect

`requeue_storable_dead_letter` selected candidates with `ORDER BY dropped_at ASC LIMIT effective_limit` (default **200**), then **skipped** rows that are no longer storable — without removing them. A row whose timestamp has aged past `EVENT_RETENTION_DAYS` can never become storable again (wall-clock only moves forward), so it held its slot in that ordering **permanently**.

Once a device carried 200 such rows, every replay cycle examined the same terminal rows and resurrected nothing. Any genuinely recoverable event dead-lettered afterwards then aged out within 7 days and was **lost for real** — on the one path whose entire purpose is to preserve it.

Measured pre-fix, with a clean control at the boundary:

```
terminal=  0 -> requeued 1
terminal=199 -> requeued 1     <- the window still fits the good event
terminal=200 -> requeued 0     <- starved
terminal=500 -> requeued 0     (examined 200, requeued 0, every cycle, forever)
```

## The fix — two mechanisms, both needed

1. **Keyset pagination** over the same ordering (not `OFFSET`, which rescans from the start each page), bounded by a scan budget, so the scan can walk *past* terminal rows to reach recoverable ones behind them. This alone moved the cliff from 200 to 20000 — better, and still a cliff.
2. **Prune** rows past retention + a 7-day margin as the scan walks them, so terminal rows cannot accumulate at the head at all. This removes the cliff: a **45,001-row backlog recovers on cycle 3** and drains completely.

## On pruning, since the DDL comments argue a blind DELETE was permanent data loss

That is true **at drop time**, when the event may still be deliverable. It stops being true once the timestamp is past retention: the server will refuse it forever, so the row preserves a *record that activity was lost*, not the activity.

The record is kept as a WARNING naming count, bucket **types** (never ids — they embed the hostname, privacy F4) and the age span. The rule is bounded strictly by timestamp age:

- an **unparseable** row is never pruned — its timestamp cannot be read, and it is the only trace of a serialisation bug;
- a **recent** row that is merely bucketless or over-long is never pruned — both are repairable by a code change.

Verified: of three unstorable rows, only the past-retention one is pruned (`{nobucket-1, overlong-1}` survive).

## Verification

**Proof of failure**: `tests/test_dead_letter_replay_starvation.py` written first — **4 failed** (200/500/1000, repeated-cycles), **2 passed** (0/199, pinning correct behaviour). All 7 pass post-fix.

**Mutation matrix — 2 mechanisms, distinct named killers, clean control:**

```
scan budget shrunk to one page  -> 5 tests redden (incl. convergence)
pruning disabled, keyset kept   -> ONLY TestBacklogConverges reddens
```

That separation is the point: the keyset scan carries the ≤budget case, pruning carries the beyond-budget case. Neither is redundant.

**Full suite: 2058 passed, 1 skipped, exit 0.**

Two existing tests legitimately changed and were updated to pin the **new** behaviour explicitly rather than loosened — the remaining-set assertion and the telemetry test's stale row.

## Not in scope

A windowed `dead_letter_dropped_24h` figure (self-clearing, so an alert on it can fire *and* clear) is the better alertable metric, but it needs a server-side column and is a separate change.

### DIFF
--- a/src/sync/queue.py
+++ b/src/sync/queue.py
@@ -1048,6 +1048,58 @@ class OfflineQueue:
     # than the bad-deploy windows seen in the incident history (~1-2h would still
     # get a retry every cooldown until it succeeds or ages out of retention).
     _DEAD_LETTER_REPLAY_COOLDOWN_SECONDS = 1800  # 30 min
+    # Replay scan budget. Terminal (past-retention) rows are skipped rather than
+    # removed, so the scan has to be able to walk PAST them to reach recoverable
+    # rows behind them — a fixed one-page window let 200 of them starve the
+    # replay permanently. Page size keeps each query small; the max bounds
+    # per-cycle cost on a large table. Parsing is the cost, ~1.5ms per 1000 rows.
+    _REPLAY_SCAN_PAGE = 500
+    _REPLAY_MAX_SCAN = 20000
+    # Extra margin BEYOND the retention window before a terminal row is pruned.
+    # At retention+margin the server has refused the event for a fortnight and
+    # always will, so the row no longer preserves recoverable ACTIVITY - only
+    # the fact that some was lost, which the prune WARNING keeps. The margin
+    # exists so anything a human might still be asked to inspect survives well
+    # past the point it stopped being deliverable.
+    _REPLAY_PRUNE_MARGIN_DAYS = 7
+
+    def _prune_terminal_rows(self, cursor, prune_ids, prune_stamps, prune_buckets) -> int:
+        """Delete dead letters that are provably past ANY use, loudly.
+
+        Called only from the replay scan, inside its transaction. A row here has
+        a timestamp older than retention + ``_REPLAY_PRUNE_MARGIN_DAYS``, so the
+        server has refused it for a fortnight and always will.
+
+        The DDL comments argue a blind DELETE of dead letters was permanent data
+        loss, and that is right at DROP time — the event might still be
+        deliverable. It stops being true once the timestamp is past retention:
+        the row then preserves a RECORD that activity was lost, not the activity.
+        This keeps that record as a WARNING naming the count, the bucket types
+        and the age span, which is what separates it from the blind DELETE.
+
+        Bucket TYPES only, never ids — a bucket id embeds the hostname, and this
+        line goes to a log that is uploadable on admin request (privacy F4).
+        """
+        if not prune_ids:
+            return 0
+        placeholders = ",".join("?" * len(prune_ids))
+        cursor.execute(
+            f"DELETE FROM dead_letter_events WHERE id IN ({placeholders})",
+            prune_ids,
+        )
+        prune_stamps.sort()
+        span = (
+            f"{prune_stamps[0]}..{prune_stamps[-1]}" if prune_stamps else "unknown"
+        )
+        logger.warning(
+            "Pruned %d dead-lettered event(s) the server can never accept "
+            "(older than retention + %dd). This is activity that was captured "
+            "and never delivered; the events are gone, this line is the record. "
+            "buckets=%s span=%s",
+            len(prune_ids), self._REPLAY_PRUNE_MARGIN_DAYS,
+            ",".join(sorted(prune_buckets)) or "unknown", span,
+        )
+        return len(prune_ids)
 
     def requeue_storable_dead_letter(
         self,
@@ -1124,6 +1176,7 @@ class OfflineQueue:
         if min_dead_age_seconds is None:
             min_dead_age_seconds = self._DEAD_LETTER_REPLAY_COOLDOWN_SECONDS
         stale_cutoff = now - timedelta(days=stale_after_days)
+        prune_cutoff = stale_cutoff - timedelta(days=self._REPLAY_PRUNE_MARGIN_DAYS)
         cooldown_cutoff = (now - timedelta(seconds=min_dead_age_seconds)).isoformat()
         requeued_at = now.isoformat()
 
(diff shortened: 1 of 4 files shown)


tudorsss:
## Reworked after review: MARK terminal rows, never delete them

The adversarial review found a **Critical in my fix**, and it was right. Reproduced independently before acting.

### The Critical

The prune fired on `not _timestamp_within(ts, prune_cutoff)`. That helper returns `False` for a **missing, null, wrong-typed or unparseable** timestamp as well as for a genuinely old one — so *"provably ancient"* and *"age unreadable"* collapsed into one branch, and the unreadable one was **deleted**:

```
PRUNED   timestamp absent
PRUNED   "timestamp": None
PRUNED   "timestamp": 1756641600          <- epoch int, i.e. a serialisation bug
PRUNED   "timestamp": ""
PRUNED   "timestamp": "Sun, 31 Aug 2026 11:00:00 GMT"
PRUNED   "timestamp": {"$date": "..."}    <- captured ONE HOUR earlier
```

My own comment three lines above asserted the opposite invariant — *"an unparseable row is NEVER pruned"*. That was true only of `json.loads` failure (whole-payload corruption), while the far likelier shape of a serialisation bug is **valid JSON with a wrong-typed field**, which it deleted. The protection was real and pointed one door along.

### The rework

Deletion was never load-bearing — pagination alone already handles 20,000 terminal rows, and the requirement was *"take terminal rows out of the candidate ordering"*, not *"destroy them"*. So:

- `dead_letter_events` gains `terminal INTEGER NOT NULL DEFAULT 0` (additive, PRAGMA-guarded migration); the candidate query adds `AND terminal = 0`. **Nothing is deleted.**
- The predicate is now **positive**: parse, require the parse to succeed, then require it to be old. `_parse_timestamp` is split out and `_timestamp_within` delegates to it, so the two cannot drift — the split exists precisely because conflating *"is it fresh?"* with *"how old is it?"* is what deleted data.

A mis-marked row costs one missed replay. A wrongly-deleted row costs the activity. Those are not the same mistake.

### Also closed from the same review

- **`stale_after_days` could move the irreversible prune.** The mark is anchored to `EVENT_RETENTION_DAYS` and cannot be widened by a caller.
- **Deleting walked `dead_letter_count` back to zero**, so a device that permanently lost activity reported *recovery* on the fleet dashboard. Marking keeps the row counted — the only fleet-visible record there is.
- **Hostname scrub leaked** on underscored machine names (`DESKTOP_ABC123`); no bucket ids are logged at all now.
- **Unchunked DELETE could reach 20,000 SQL variables** (fails below SQLite 3.32); the UPDATE is chunked at 500.
- **`pruned` missing from two early returns**, so a test read it unguarded and `KeyError`d instead of failing with its message.

### The test gap, which was the sharpest finding

Every fixture in the new file built events through one helper that always emitted `ts.isoformat()` — so the only two inputs it could produce were the two where a correct rule and the broken one **agree**. The reviewer predicted a mutant marking every unstorable dict would be invisible. Measured, that was true.

`TestUnreadableTimestampsAreNeverMarked` closes it: seven unreadable shapes asserted to survive *and* stay unmarked, a positive control (a provably-30-day row **is** marked, else the class passes for a rule that never marks anything), and a test that a caller-supplied window cannot widen the mark.

```
M-blind  (mark every unstorable dict)  -> 7 named tests redden   [was: invisible]
M-off    (never mark)                  -> convergence + positive control
M-window (caller widens the mark)      -> the caller-window test
control  clean, tree clean
```

### One more signal worth recording

Both existing tests I had edited to accommodate deletion are **restored to their `origin/main` assertions and pass unmodified**. Deletion required changing them; marking does not, because it preserves the semantics they already encoded. That is a better argument for the approach than anything I wrote above.

**Full suite: 2068 passed, 1 skipped, exit 0.**

### Still open from the review, deliberately

- **Recent-but-unstorable rows** (no `bucket_id`, over-long duration) keep a recent timestamp and are not marked, so more than 20,000 of them could still occupy the scan budget for up to 14 days. That is the original bug at 100x the trigger threshold, self-healing. Reachable via `backfill_status_bucket_ids`, which rewrites `queued_events` only. Worth a durable scan watermark; not this PR.
- **`headroom <= 0` returns before marking**, so a device whose live queue is at capacity gets neither replay nor marking — precisely when the backlog is largest.
- **The 7-day margin assumes the server's retention is ≤ 7 days.** `EVENT_RETENTION_DAYS` mirrors a constant in another repo that I have not read. This now costs a missed replay rather than data, which is why it is no longer a blocker.

tudorsss:
## Refutation review: no Critical. Two Important closed, one Minor closed, one accepted.

### Important 1 — the exception defeated itself

Rows the rule *refused* to mark (missing/unparseable timestamp, non-dict payload) were never storable either, so they held their slots in the `ORDER BY dropped_at` ordering and starved the scan at `_REPLAY_MAX_SCAN`. The original defect, reproduced at 20,000 instead of 200, and it never self-cleared:

```
before:  cycle 0/1/2 -> examined 20000, requeued 0, marked_terminal 0
after:   cycle 0 -> marked 20000
         cycle 1 -> requeued 1        (20,001 rows RETAINED, nothing deleted)
```

The reviewer's framing is the part worth keeping: the exception existed to protect *"the likely shape of a serialisation bug"*, and a serialisation bug is **systematic** — it produces that shape in **bulk**, which is the only volume at which starvation happens. The shape the exception protected was the shape that defeated it.

**This is a correction to my own reasoning.** I refused to mark those rows because I could not prove their *age*. Permanence never required age: `is_event_storable` demands a timestamp that parses, so a row whose timestamp cannot be read can never be storable, and dead-letter payloads are immutable. That caution was inherited from the DELETE design, where it was correct and prevented real data loss. Under MARKING it was the wrong caution — the row is retained, still returned by `get_dead_letter_events`, still counted by `dead_letter_count`. A serialisation bug stays exactly as diagnosable and stops blocking every event behind it.

### Important 2 — the suite structurally could not see it

No test composed *"more than the scan budget of **unmarkable** rows"* with a convergence check. Every convergence test used a markable class; the unreadable-timestamp tests used a single row. `TestUnmarkableBacklogStillConverges` is that composition.

`TestUnreadableTimestampsAreNeverMarked` is renamed to `...AreNeverDeleted` and asserts the property that actually matters — retained, inspectable, payload intact. Asserting *"not marked"* was the wrong invariant and is precisely what left those rows starving the scan.

### Minor 3 — fixed

The bool guard on `duration` returned from the whole function, so a boolean duration exempted a provably-ancient row from marking. It now skips only the duration branch, witnessed by a test.

### Minor 4 — accepted, not fixed

`skipped_unstorable` overcounts by up to one page when the inner loop breaks at the requeue limit. Telemetry only; no consumer reads the dict (`sync_engine.py:3715` discards it).

### Attacks that came back clean

Immutability premise (only writes are two INSERTs, the terminal UPDATE, the replay DELETE — the bucket backfill and project-id sanitiser touch `queued_events` only); storable-and-permanently-unstorable is unreachable; the `ALTER TABLE ... NOT NULL DEFAULT 0` migration is legal on every SQLite in play and downgrade-safe (both INSERT sites use explicit column lists); keyset `(dropped_at, id)` ordering is correct with duplicate timestamps and needs SQLite ≥ 3.15 against 3.45.1 local; `headroom <= 0` is self-clearing and loses nothing.

### Mutation matrix — distinct named killer each, clean control

```
unreadable timestamp not marked (revert)  -> TestUnmarkableBacklogStillConverges
bool duration returns from function       -> test_a_boolean_duration_does_not_exempt_an_ancient_row
mark everything unstorable                -> test_the_caller_window_cannot_widen_the_mark
```

**Full suite: 2072 passed, 1 skipped, exit 0.**

### Not chased

Whether the shipped PyInstaller Python is ≥ 3.11. Below that, `datetime.fromisoformat` rejects several valid ISO-8601 spellings, which would be a second bulk source of unreadable timestamps. That class is now marked rather than left to starve the scan, so it degrades to "those rows are retained and skipped" instead of a starvation — worth a check of the build matrix regardless.

PR #246, opened 2026-09-03
fix(aw): a held socket is not a capture — one rule at every attach point

tudorsss:
## The class

`_server_responding()` was extracted by `#213` *"so the two callers cannot drift"*. The probe did not drift; the **carve-outs around it** did. "Is an external server actually capturing on our port" ended up written at four attach points in `_start_locked` plus the recovery predicate in `restart_if_needed`, and the copies disagreed:

| site | asks `/api/0/info`? | re-latches? | |
|---|---|---|---|
| `:1243` Rosetta attach | yes | yes | correct — this is the reference |
| `:1320` backoff attach (`#230`) | yes | **no** | F-1 |
| `:1354` download-failure attach | **no** | n/a | F-2 |
| `:1373` normal-path attach | **no** | n/a | F-3 |
| `:1702` external-vanished recovery | **no** | n/a | F-4 |

Every defect is one of those copies missing a half the others have, so the fix is one helper — `_external_server_capturing()` — used at all five.

## The four defects

All measured on unmodified `main`.

**F-1 — `#230` clears the capture-dead flag and never re-latches it.** The clear was copied from the Rosetta sibling and the `else` was not, so the clear from a *previous cycle of that same branch* survives into one where nothing is capturing:

```
cycle 1  external server ALIVE   started=True   download_failed=False
cycle 2  external server DEAD    started=False  download_failed=False
```

Control — the Rosetta sibling forty lines up, same probe:

```
cycle 2  external server DEAD    started=False  download_failed=True   <- RE-LATCHES
```

So a device capturing nothing published `tracker_download_failed=False`, which is what the fleet's no-capture alert reads, for up to `DOWNLOAD_RETRY_MAX_INTERVAL` (an hour) per cycle — on exactly the escalated-backoff device `#230` was written for. Pre-`#230` it reported a false capture-*dead*; that traded it for a false all-clear on the same field, which is the worse direction.

**F-2 — `#233` says "the last two sites" and there was a third.**

```
download fails, socket held but dead -> started=True  using_external=True
                                        _server_responding never consulted
```

**F-3 — the widest, and in neither review.** It needs no download failure and no Rosetta. With good binaries on disk, a port held by a process that answers nothing made us defer to it and run our watchers into a server that answers nothing — observably identical to attaching to a healthy shared ActivityWatch:

```
healthy external AW server   started=True  using_external=True  our_components=2  server_asked=False
port held by a DEAD process  started=True  using_external=True  our_components=2  server_asked=False
port free                    started=True  using_external=False our_components=3
```

**F-4 — why F-3 was permanent.** The "external server vanished" recovery keyed on the port being *released*, and a corpse holds it forever:

```
_using_external=True   _port_in_use()=True
recovery condition (_using_external and not _port_in_use) = False   <- WEDGED
```

## Proof of failure

Four defect tests fail pre-fix; six correct-behaviour tests already pass, so the suite isolates the bugs rather than describing the new code.

Mutation matrix — control 10 passed, every mutant collects all 10 (no crash), each mechanism reddens a **distinct** named test:

| mutant | reddens |
|---|---|
| helper stops asking (bare TCP) | 4 — the held-but-dead test at every site |
| F-1 re-latch removed | 1 — `test_flag_re_latches_when_the_external_server_dies` |
| F-4 predicate back to bare port | 1 — `test_recovery_fires_when_the_holder_stops_answering` |
| **helper always refuses** (allowance witness) | 7, incl. all four live-server controls |

That last row is the one that matters for blast radius: a guard like this fails dangerously in *both* directions, and the over-refusal mutant proves users running their own ActivityWatch cannot be silently cut off.

Full suite: **2110 passed, 1 skipped** (baseline 2100 + the 10 added).

## Three existing tests were updated, not weakened

They stub `_port_in_use` and never stubbed `_server_responding`, because a held socket was all the code read. `test_aw_manager_external_selfheal.py` says `# external server healthy` on that very line. They always *meant* a live server; they now say so.

That is the agreement region this whole class lived in — **no fixture could express the difference between a live server and a corpse, so no test could have caught any of these.** The `#230` suite is the sharpest case: every fixture hard-sets `tracker_download_failed = True`, so nothing ever drove that branch twice, which is the only way to see F-1.

## Machine-dependence check

The change makes `_server_responding` reachable from more paths, so I checked whether the suite now depends on whether ActivityWatch happens to be running locally. With the real method made to raise, exactly **one** test reaches it suite-wide — the pre-existing `test_server_responding_is_the_http_ask_wait_for_server_polls`, whose subject it is. This adds no live HTTP call.

## Scope

F-1 and F-2 are **live in v1.5.130**. F-3 and F-4 predate it.

Found while auditing `#217`/`#222`/`#229`/`#230`/`#233`/`#240`/`#241`; F-3 and F-4 were not in either review pass and came out of tracing why F-2's wedge could never self-heal.

https://claude.ai/code/session_01X89WdSQtzR64U7tkAsjxDN

### DIFF
--- a/src/aw_manager.py
+++ b/src/aw_manager.py
@@ -1240,7 +1240,7 @@ class AWManager:
             # is, in fact, running an ARM ActivityWatch which just failed to
             # answer /info within 2s — visible, wrong for one cycle, and it
             # self-corrects on the next probe.
-            if server_already_running and self._server_responding():
+            if self._external_server_capturing():
                 # ...but an external server IS capturing on this port, so this
                 # is not a "capturing nothing" device. Verbatim the rule the
                 # download-failure branch below already applies: attach, log,
@@ -1317,7 +1317,7 @@ class AWManager:
                     self._download_retry_interval,
                 )
                 self._managed_components_unavailable = True
-                if server_already_running and self._server_responding():
+                if self._external_server_capturing():
                     # The comment above says "an attached external server still
                     # captures" and this branch did not carry it out: it left
                     # tracker_download_failed latched and _using_external False,
@@ -1334,7 +1334,18 @@ class AWManager:
                     self._using_external = True
                     self.tracker_download_failed = False
                     return True
-                return server_already_running
+                # RE-LATCH, symmetric with the clear above and with the Rosetta
+                # sibling's `else` forty lines up. #230 copied the clear and not
+                # this, so the clear from a PREVIOUS cycle of this same branch
+                # survived into one where nothing is capturing: the device kept
+                # publishing tracker_download_failed=False while recording
+                # nothing, for up to DOWNLOAD_RETRY_MAX_INTERVAL (an hour) per
+                # cycle, which the fleet's no-capture alert reads as healthy.
+                # Its own tests could not see it -- every fixture hard-set the
+                # flag True, so no test drove this branch twice.
+                self._using_external = False
+                self.tracker_download_failed = True
+                return False
             self._last_download_attempt = now
             logger.info("Tracker components not found, downloading...")
             install_dir = _get_install_dir()
@@ -1351,7 +1362,14 @@ class AWManager:
                     self._download_retry_interval * 2, DOWNLOAD_RETRY_MAX_INTERVAL
                 )
                 self._managed_components_unavailable = True
-                if server_already_running:
+                if self._external_server_capturing():
+                    # Asked, not assumed. This branch read the bare TCP connect
+                    # until #246: a socket held by a process that no longer
+                    # answers /api/0/info reported this device as started and
+                    # attached while it captured nothing. #233 closed this class
+                    # at two sites and its message said "the last two"; this was
+                    # the third.
+                    #
                     # An external server is already capturing on this port, so
                     # this is NOT a "capturing nothing" situation — only our
                     # managed watchers are unavailable. Log it; don't latch the
@@ -1370,13 +1388,27 @@ class AWManager:
                 self._dispatch_download_failure_report()
                 return False
 
-        if server_already_running:
+        if self._external_server_capturing():
             logger.info(
                 f"Tracker server already running on port {self.aw_port}, "
                 "using external instance"
             )
             self._using_external = True
         else:
+            if server_already_running:
+                # Held, but not answering. The widest member of this class and
+                # the only one needing no download failure and no Rosetta: with
+                # perfectly good binaries on disk we deferred to whatever owned
+                # the port, then ran our watchers into a server that answers
+                # nothing. Starting our own is the same thing we do when the
+                # port is free; if the corpse still holds the bind we fail
+                # loudly here instead of reporting a healthy attach.
+                logger.warning(
+                    "Port %s is held by a process that does not answer "
+                    "/api/0/info — starting our own trackers rather than "
+                    "attaching to it",
+                    self.aw_port,
+                )
             logger.info(f"Starting tracker components from {binaries_dir}")
 
             # Start server first
@@ -1699,8 +1731,11 @@ class AWManager:
         # special — otherwise fall through to watcher health/stale recovery,
         # which MUST run in external mode too. Returning early there was why a
         # blind/orphaned bf-idle-tracker never self-healed after an update.
-        if self._using_external and not self._port_in_use():
-            logger.warning("External tracker server no longer running — starting our own")
+        if self._using_external and not self._external_server_capturing():
+            logger.warning(
+                "External tracker server no longer answering /api/0/info — "
+                "starting our own"
+            )
             self._using_external = False
             return self._start_locked()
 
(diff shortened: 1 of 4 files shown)


tudorsss:
## Gate round 1 — a Critical against this branch, from the correctness lens. Fixed in `d2facc1`.

Both review lenses independently landed on the same area, and the correctness lens filed a Critical **against this PR's own fix**. It reproduces. Control reverting only the F-4 predicate, everything else identical:

```
asked /info   rc=False  _using_external=False  watchers=[]   is_managing=False
bare port     rc=True   _using_external=True   watchers=[2]  is_managing=True
```

On a **healthy** shared ActivityWatch that missed two `/api/0/info` answers — a load spike, a wake from sleep, a long AW query — the new predicate dropped external mode, spawned our own server against the port that healthy server still owned, failed to bind, and `_start_locked`'s blanket `self.stop()` cleared every watcher. `is_managing` then reads `False`, so `main.py`'s 60s tick stops calling `restart_if_needed` and nothing re-enters the branch. Permanent, silent, on a machine that was fine — with **both** capture-dead flags reading `False`, which is the exact reassuring-direction failure this branch exists to close.

Strictly worse than the bug it was fixing.

### What changed

1. **F-4 reverted** to the bare port read, with the measurement recorded at the line and a test pinning the deferral so nobody flips it back. Doing it properly needs a debounce (N consecutive non-answers, resetting on any success) — its own change with its own evidence. Its value was recovering a device already attached to a corpse; F-3 stops that state being entered, and an existing one recovers on the next app start.
2. **A failed bind while the port is HELD no longer tears down.** The holder owning the port is evidence *about the holder*, never a reason to destroy working watchers. F-3's own failure path needs this regardless of F-4. It also closes what the design lens raised separately (M-1): that arm was the only member of this class latching no flag at all, so the site this PR calls widest published a reassuring pair on a device capturing nothing.
3. **I-4** — `test_a_held_but_dead_port_does_not_clear_the_flag` now asserts the return as well as the flag. Verified the mutant surviving on `fe98c50` (2110 passed) and killed on `d2facc1`.
4. **Two stale rationales** updated — the Rosetta branch's "Here — and ONLY here", already half-false after `#233` and fully false now. Kept the half still true and specific.

### Matrix after the amendment (control 14 passed, all mutants collect 14)

| mutant | reddens |
|---|---|
| helper stops asking | 3 |
| helper always refuses (allowance) | 6, incl. all live-server controls |
| F-1 re-latch removed | 2 — incl. the previously-surviving return |
| teardown guard removed (the C-1 regression) | 1 |
| teardown guard always fires | 1 — the free-port control |

The last two are the two ends of the new guard, both witnessed.

Suite **2114 passed, 1 skipped** (baseline 2100 + 14). F-1/F-2/F-3 re-verified unchanged after the amendment.

### Deferred, deliberately

- **F-4 with a debounce** — pinned by `TestF4RecoveryStillReadsTheBarePort_Deliberately`.
- **The sixth site** the design lens found at `aw_manager.py:1774` (`reachable = self._port_in_use()` feeding `_is_window_tracker_stale`). Same class, opposite consequence — a corpse manufactures a false *stale* verdict rather than a false all-clear. Pre-existing, outside this PR's declared scope, worth its own change.

tudorsss:
## Gate round 2 refuted round 1's repair. Scope cut in `4dfc290`.

Round 2 filed a Critical against the fix from round 1, and it reproduces. Same probe, `main.py`'s real tick order, a corpse on `:5600` that releases the port at tick 4:

```
origin/main   tick1-3 WATCHERS=2 ...  tick4: 3 procs   FULLY RECOVERS
d2facc1       tick1-5 WATCHERS=0 ...                   never recovers
```

**Mechanism.** `_start_component(BF_SERVER)` succeeds — Popen returns — *before* `_wait_for_server` fails, so `_processes` holds a dead `bf-data-service`. My guard skipped `self.stop()` to preserve watchers. But emptying `_processes` is exactly what re-arms `set_capture_suppressed`'s `elif not self._processes` rebuild route (`aw_manager.py:752`), which is the **only** path back into `_start_locked`. Preserving the watchers disarmed the thing that would have rebuilt them.

Scorecard for this branch's own repairs:

| round | repair | result |
|---|---|---|
| 1 | F-4 asks `/info` | healthy machine loses all watchers on a 2-answer blip, permanently |
| 2 | skip the `stop()` | zero watchers permanently **and** the rebuild route disarmed — worse than `origin/main` |

Two rounds, two Criticals, both caused by my own repairs, in the same place. Per `coding-standards.md` that is a design defect rather than fix-forward, and per `fail-closed.md` Rule 18 it means the property is unreachable at this layer. So I stopped repairing and cut the scope instead of attempting a third.

### What ships now

Only the parts that never drew a finding, and that return **before any spawn** — so they cannot touch the teardown path that produced every Critical here:

- `_external_server_capturing()`, applied at the Rosetta attach and the two download-path attaches
- **F-1** — the backoff attach re-latches: `cycle 2 (server dead) → download_failed=True`
- **F-2** — the download-failure attach asks: `started=False, download_failed=True, asked=True`

Both are live defects in v1.5.130 and both published a reassuring flag pair on a device capturing nothing.

### What is deferred, and why it is not a TODO

**F-3** (normal-path attach) and **F-4** (recovery predicate) are reverted. Both are still real bugs. Neither is fixed here. The measurements are recorded at both lines and pinned by tests, so flipping either line alone reddens the suite — verified: `M4 → test_a_dead_holder_is_still_attached_to_for_now`.

The fix belongs at a different layer: either the rebuild route stops being gated on `_processes` being empty, or the spawn reaps only the server it started. The blanket `self.stop()` also gained a comment saying it is load-bearing, because "preserve the watchers" is the obvious next repair and it is the one that fails.

### Verification

Corpse scenario now behaves identically to `origin/main` — watchers survive, full recovery on release. Matrix, control 11 passed, all mutants collect 11:

| mutant | reddens |
|---|---|
| helper stops asking | 2 |
| helper always refuses (allowance) | 4, incl. both live-server controls |
| F-1 re-latch removed | 2 |
| F-3 flipped to ask (deferral pin) | 1 |

Suite **2111 passed, 1 skipped** (baseline 2100 + 11). Round 3 running against the reduced diff.

PR #252, opened 2026-09-07
fix(errors): send the fingerprint so the ingest groups on it

### LINKED ISSUE #251: Error reports never carry their fingerprint, so one recurring fault splits into many in the tracker
## Steps to reproduce

Not UI-reachable — this is the agent's ops error-reporting path. Run this against a checkout at current `main`:

    cd ~/Code2/betterflow-sync
    python3 -c "
    import sys; sys.path.insert(0,'src')
    from error_reporter import ErrorReporter
    r = ErrorReporter.__new__(ErrorReporter)
    r._environment='prod'; r._release='1.5.132'; r._context_provider=None
    p = r._build_payload('Dropped 1 queued event(s)', 'warning', None, {}, {})
    print(sorted(p.keys()))
    "

1. Run the command above.
2. Read the printed key list.
3. Observe that `fingerprint` is absent, even though every caller in this repo passes one.

## Actual result

    ['context', 'environment', 'level', 'message', 'project', 'release', 'tags']

No `fingerprint` key. `capture()` accepts `fingerprint=` and uses it **only** as an in-process cooldown key (`src/error_reporter.py:123`); `_build_payload` (`src/error_reporter.py:210-219`) never copies it into the POST body. So every deliberate fingerprint in this repo — `offline-queue-events-dropped`, `sync-watchdog-timeout`, `rosetta-notice-undelivered`, the watchdog duration bands — dies on the machine.

Observed downstream on 2026-09-07: the message **"Dropped 1 queued event(s) after max retries — the server rejected them"** appears in the tracker as **five separate open issues with five distinct fingerprints** (`ec601751`, `5ffc28ee`, `0446cbec`, `6786f186`, `1c6a276c`) over two days, despite identical leading text and an identical intended fingerprint.

## Expected result

One recurring fault is one row in the tracker, with an occurrence count. The fingerprint the agent chose is the fingerprint the tracker groups on.

## Technical details

**Verified** (read in this repo, `main` at `42f34c0`):

- `src/error_reporter.py:123` — `key = fingerprint or f"{level}:{message}"`, then `_should_send(key, ...)`. Local dedup only.
- `src/error_reporter.py:210-219` — the payload dict. Seven keys, plus `stack` when an exception is passed. No fingerprint.
- Callers that supply one and are therefore affected: `src/sync/sync_engine.py:4167` and `:4176`, `src/main.py:1781`, `:1794`, `:2003`, and others.

**Verified** (read in `Better-Quality-Assurance/betterqa-bot` on its default branch, 2026-09-07 — cross-repo, so this describes that repo today and not a branch of it):

- `src/monitor/error-fingerprint.ts:63` `generateFingerprint()`. With no client fingerprint and no stack frames it falls through to `normalizeMessageForFingerprint(opts.message)` (`:86-90`) — the first 200 chars, number- and quote-normalized, djb2-hashed.
- `src/webhooks/error-ingest.ts:843` reads `error.fingerprint` and honours it verbatim when the credential is trusted; `:470-474` rates a `DSN` credential `"trusted"`. The agent authenticates by DSN (`src/error_reporter.py` `_post`, `Authorization: Bearer {self._dsn}`), **so a fingerprint we send would be honoured** — the field is simply never sent.

Why the split is five and not one: our drop message embeds `buckets=<types>`, a window-age phrase and a reason code. Number-normalization collapses `164.8s` in the watchdog message (which is why the overrun issue stays a single fingerprint at 2x), but it does not collapse differing bucket type names or reason codes.

**Second-order consequence worth fixing alongside:** the reasoning at `src/main.py:2004-2018` argues about how the server-side digest counts each duration band. That reasoning is currently void — the bands never reach the server, so `dedup_window=0` changes only local throttling.

**Why this was not caught.** About ten tests assert the fingerprint kwarg was *passed to a mocked* `capture()` — `tests/test_sync_engine.py:1181`, `tests/test_watchdog_overrun_outcome.py:64`, `tests/test_afk_convergence_hardening.py:148` and others. None assert it reaches the payload. That is `test-fixture-discipline.md` Phantom 7: the tests assert the producer, and the consumer (the wire payload) is unguarded, so the feature can be silently absent with the whole suite green. Any fix needs an assertion on `_build_payload`'s **output**, not on the mock.

## Environment

- Repo: `betterflow-sync`, `main` at `42f34c0`, `__version__` 1.5.132.
- Observed against the live ops ingest (betterqa-bot `/notify/error`), items open as of 2026-09-07.
- Affects every platform and every released version that has the current `_build_payload` — the field has never been sent.

## Business impact

Ops cannot tell a recurring fault from a burst of new ones: one agent problem hitting five machines, or one machine five times, arrives as five separate unresolved items to triage individually. The fleet's own count-based signals are the thing being distorted, so anyone reading the tracker to judge how widespread a problem is gets an inflated issue count and a deflated per-issue occurrence count. It also means the deliberate grouping decisions written throughout the agent — including the watchdog duration bands built specifically so their counts would publish a distribution — have never actually taken effect.

tudorsss:
Closes #251.

## What was wrong

`capture()` accepted `fingerprint=` and used it **only** as an in-process cooldown key (`src/error_reporter.py:123`). `_build_payload` never copied it into the POST body, so no fingerprint this agent has ever chosen reached the ingest — `offline-queue-events-dropped`, `sync-watchdog-timeout`, the watchdog duration bands, all of it died on the machine.

The ingest then falls back to hashing a normalized message, which is wrong in **both directions at once**:

- normalization eats the elapsed figure, so the three duration bands collapse into one row;
- it leaves bare identifiers alone, so one recurring drop fault split across **five** open tracker rows, because that message embeds `buckets=<types>` and a reason code.

Counts were always correct — it is the grouping that was wrong. Worth saying plainly so nobody re-derives fleet numbers that were fine.

## The change

One field, sent only when the caller supplies one:

```python
if fingerprint:
    payload["fingerprint"] = fingerprint[:_MAX_FINGERPRINT]
```

Omitting it when absent is deliberate. A missing field means "hash it yourself" at the ingest, which is exactly what those callers already get; sending the local `level:message` fallback would silently change grouping for every caller that never asked for one. The 128-char cap matches what the ingest truncates to, so two keys can't differ only past the cut and group together with nothing local explaining why.

The agent authenticates by DSN, which the ingest rates `trusted`, so a supplied fingerprint is honoured verbatim rather than hashed. (Read on `betterqa-bot`'s default branch 2026-09-07 — a statement about that repo today, not about a branch of it.)

## Proof of failure

The two defect-specific assertions were run against the pre-fix file:

```
E   KeyError: 'fingerprint'   test_explicit_fingerprint_is_sent
E   KeyError: 'fingerprint'   test_fingerprint_survives_verbatim_when_the_message_would_normalize_away
2 failed, 1 passed
```

The third (`no fingerprint -> no field`) **passed pre-fix** and pins the behaviour that must not change — only the defect cases fail, which is the shape that shows the tests isolate the bug.

Post-fix: 28/28 in that file; **2155 passed / 1 skipped** across the suite, against **2153 collected** on `main` — up by exactly the three tests added. `ruff` clean on both touched files.

## Why this hid

About ten existing assertions check the fingerprint kwarg against a **mocked** `capture()`. That is the producer side, and it stayed green for the entire life of the defect. The new tests assert the **posted body** instead. Any future work here should do the same.

## Heads-up for whoever watches the tracker

When this ships the rows move in both directions on the same day:

| | before | after |
|---|---|---|
| Sync overrun | 1 fingerprint | up to 3 (one per band); old count -> 0 |
| Dropped queued events | 5 fingerprints | 1, with a real recurrence count |

Neither is a change in fleet behaviour — same cycles, same drops, grouped as intended. The **split is the one that looks like a regression**: it presents as new fingerprints appearing, so alerting keyed on "a new fingerprint showed up" can trip for no real-world reason. Report volume is unaffected; `dedup_window=0` on the overrun path is a client-side throttle and stays client-side.

Worth landing while somebody is watching rather than quietly.


tudorsss:
## Pre-merge review gate: both lenses run, findings fixed

**Correctness lens — no Critical, no Important.** It verified the premise I had only read: the ingest gates custom fingerprints on a credential trust tier this PR never mentioned (`error-ingest.ts` `CUSTOM_FINGERPRINT_TRUST`), and on the publishable tier `acceptCustomFingerprint` **silently discards** the field with a 200 — this PR would have been a no-op no local test could detect. The agent authenticates by DSN, which maps `trusted`, so it holds. It also reproduced the handshake with named mutants and confirmed the 128 figure is the ingest's own `slice(0, 128)`, with no narrower DB cap behind it.

**Design lens — design sound, two Importants, both now fixed in `bf25e2a`.**

**1. The cap desynced the two keys.** `capture()` used the untruncated fingerprint as the local cooldown key while the wire got `[:128]`. Two keys differing only past the cap would take separate local cooldowns and land in one ingest group — making the docstring's "used for BOTH" quietly false. The local key is now truncated the same way. The `level:message` fallback stays uncapped deliberately: it never leaves the machine, and capping it would collapse every long message into one cooldown key.

**2. The cap had no witness**, flagged by both lenses and measured by one (deleting the slice left the suite at 28 passed). Now pinned by property rather than by constant — two fingerprints identical through 128 chars are one group at the ingest, so they must share one local cooldown:

```
control                29 passed
M1 wire cap removed  -> 1 failed  test_cap_is_applied_and_keeps_the_two_keys_in_step
M2 local cap removed -> 1 failed  (same test, different assertion inside it)
```

**3. "The ingest hashes a normalized message" was true only by an accident of language.** It prefers a normalized top *stack frame* and reaches the message only when it cannot extract one — and its extractor keeps lines beginning `at `, a JS shape a Python traceback (`File "...", line 42, in foo`) never matches. Three sites here **do** send a stack (both excepthooks, `main.py`'s crash report), so a regex over there teaching that extractor about Python frames — a pure improvement in that repo, with no reason to consider this client — would silently regroup all three. The precondition is now stated at both the comment and the docstring, so the next reader re-checks the mechanism rather than the outcome.

Post-fix: **2156 passed / 1 skipped**, `ruff` clean on both files.

## Sharpening the rollout note above

The correctness lens added a detail worth having before this lands: because the row key is `(project, fingerprint)`, every fingerprinted call site mints a **new** `error_reports` row on its first post-upgrade occurrence — `occurrences=1`, `isNew: true`, so **one alert each** — while the old message-hash row goes quiet and keeps its history. Expected, and it will look like a burst of new errors from a fleet that has not changed.

## Not fixed, deliberately (optional, per the evidence standard)

- A non-`str` fingerprint would now drop the whole report (`[:128]` raises `TypeError`, swallowed by `capture()`'s outer guard) where pre-fix the report still went. **No caller can reach it** — all 20 sites are string literals, f-strings, or a `str` tuple. Logged because the ingest went the other way on the same question after `{"fingerprint": 123}` once took a request down; if a dynamic fingerprint ever appears here, coerce.
- `if fingerprint:` vs `is not None` — unwitnessed, harmless either way (the ingest rejects an empty string).

## Not verified

Both lenses read `betterqa-bot` at a **local checkout** (`42ce05ee`), not the deployed service. A checkout is not the running artifact. The behaviour degrades safely if the deployed gate differs — an untrusted tier drops the field and the event still lands under the derived hash — so neither treated it as blocking, and neither sent live traffic to production to settle it.

PR #256, opened 2026-09-11
fix(sync): the ops alert saw only the first failure reason

tudorsss:
Follow-up to #254.

## What was wrong

#254 made the "sync failing repeatedly" ops alert name the local failure kinds instead of giving a bare count. But the production caller in `SyncCoordinator._do_sync` passed `stats.errors[0]`, a single string, so `server_status_summary` only ever saw the first reason. When that first reason is the one producer that is deliberately never named (sync_engine appends `result.error` verbatim), any nameable reason behind it was dropped and ops got the pre-#254 bare count again.

Measured against the shipped function on `origin/main` (9227939):

```
full list      -> '; 2 local reason(s) [bucket-sync-failed] recorded in local dead-letter'
errors[0] only -> '; 1 local reason(s) recorded in local dead-letter'
```

Every test in `tests/test_sync_local_reason_kind.py` passes a list. None passed what the real caller passed.

## Fix

- `_do_sync` passes `stats.errors or "Sync failed"` (the whole list).
- `_note_sync_failure` accepts `str | list[str]`. The local log line joins the list.
- The tray still shows `errors[0]`. That's one line of local UI, not the ops ingest.
- Redaction is unchanged. Emitted tokens still come only from `_LOCAL_REASON_KINDS`, or they're counts.

I grepped for other `errors[0]` callers. The only other one is the tray call on the next line, and it's deliberate. The `except` path at `main.py:1927` passes the literal `"Sync error"`.

## Corrected rationale

The `server_status_summary` comment and the docstring of `test_two_rejections_sharing_one_status_are_not_reported_as_mixed` said the pre-#254 output was `plus 0 local reason(s)`. Running the pre-#254 bytes actually gives `; server status 422 plus 1 local reason(s), full detail in local dead-letter` (2 items minus 1 de-duplicated code). `plus 0` could never happen on that branch. Both are now corrected to the measured string.

## Proof

- New `tests/test_sync_failure_alert_names_every_reason.py` drives `_do_sync`, not the helper, with the unnamed reason first and a nameable one second.
  - On pre-fix `main.py`, `test_a_nameable_reason_behind_an_unnamed_one_reaches_the_ops_alert` fails with `the nameable reason was dropped: Sync failing repeatedly (3×): 1 local reason(s) recorded in local dead-letter`.
  - The leak test and the empty-errors control already passed pre-fix, because that behaviour was already correct.
  - Post-fix, all 3 pass.
- Redaction mutant (`local_reason_kinds` returns the raw reasons instead of kinds): 8 tests fail, including the existing `test_the_interpolated_text_never_leaves_the_device` and the new caller-path leak test. Control run: 13 passed. The file was restored and the tree is clean.
- Full suite locally (CI's `pytest tests/`): 2169 passed, 1 skipped.
- `verify_tracker_pins.py`: all 4 pins verified. This diff is outside that workflow's path filter anyway.

## Out of scope (noted, not changed)

- The 300s dedup key is the fixed fingerprint `sync-repeated-failure`, so the first reason in a window wins and later, different reasons in the same window are suppressed.
- The partial-success path (`stats.success or stats.events_sent > 0` with `stats.errors` set) only logs locally and never reaches ops.

Merging this doesn't release it. The desktop agent ships through a tagged release.


tudorsss:
## Pre-merge refutation review (adversarial-reviewer, branch `fix/sync-failure-names-every-reason` at 85d95c0)

The reviewer tried to find an input where the ops alert still loses a nameable reason or leaks interpolated text, and found no Critical or Important issues.

- It checked the proof-of-failure independently: with `stats.errors[0]` restored in a scratch copy, `test_a_nameable_reason_behind_an_unnamed_one_reaches_the_ops_alert` fails with `1 local reason(s) recorded in local dead-letter`. On the new code, 13 tests pass.
- Nothing non-string can reach `stats.errors`. Every writer in `sync_engine.py` appends an f-string or a truthy `result.error`.
- Output stays bounded to status digits plus `_LOCAL_REASON_KINDS` tokens, so a longer list can't widen what leaks.

Its first optional note is fixed in 51a33df: a comment cited the stale `main.py:1898` and overstated "nothing else on the wire carries a cause".

Two older gaps it found, neither changed here:

1. `_do_sync`'s `except` branch passes `exc=e`, so the full traceback, including `str(e)`, reaches the ingest as `stack` under this same `sync-repeated-failure` fingerprint.
2. "Failed to get buckets" never alerts ops. `aw_bucket_fetch_failed` returns early to `_handle_aw_bucket_failure`, which never counts toward `_consecutive_sync_failures` or calls capture.

